PR #22, opened 2020-03-20
add graphql functions

weihou:
will be used in hosted fields

### DIFF
--- a/src/graphql.js
+++ b/src/graphql.js
@@ -3,6 +3,7 @@
 import { ZalgoPromise } from 'zalgo-promise/src';
 import { request } from 'belter/src';
 import { CURRENCY } from '@paypal/sdk-constants/src';
+import { stringifyError } from 'belter/src';
 
 import { getLogger } from './logger';
 import {
@@ -122,7 +123,7 @@ export function getGraphQLFundingEligibility<T>(fields : string) : ZalgoPromise<
         }
         return gqlResult && gqlResult.fundingEligibility;
     }).catch(err => {
-        getLogger().error(`graphql_fundingeligibility_error`, { err: err.stack || err.toString() });
+        getLogger().error(`graphql_fundingeligibility_error`, { err: stringifyError(err) });
         return ZalgoPromise.reject(err);
     });
 }
--- a/src/graphql.js
+++ b/src/graphql.js
@@ -1,0 +1,142 @@
+/* @flow */
+
+import { ZalgoPromise } from 'zalgo-promise/src';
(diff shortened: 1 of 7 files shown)


src/graphql.js
```diff
@@ -0,0 +1,142 @@
...
+                    buyerCountry:$buyerCountry,
+                    currency:$currency,
+                    intent:$intent,
+                    commit:$commit,
+                    vault:$vault,
+                    disableFunding:$disableFunding,
+                    disableCard:$disableCard,
+                    merchantId:$merchantID
+                ) {
+                    card {
```
bluepnume: Can we make

```
                    card {
                        eligible 
                        branded 
                    }
```

a parameter that is passed in by card-components, so anyone can pass in their own query?

src/graphql.js
```diff
@@ -0,0 +1,142 @@
...
+            variables
+        }).then((gqlResult) => {
+            if (!gqlResult || !gqlResult.fundingEligibility) {
+                throw new Error(`GraphQL fundingEligibility returned no fundingEligibility object`);
+            }
+            return gqlResult && gqlResult.fundingEligibility;
+        });
+    }
+    catch (err) {
+        getLogger().error(`GraphQL fundingEligibility error`, err);
```
bluepnume: Let's do `graphql_fundingeligibility_error` -- error name needs to be a cal event, and payload should be an object like `{ err: stringifyError(err) }`

src/graphql.js
```diff
@@ -0,0 +1,142 @@
...
+            `,
+            variables
+        }).then((gqlResult) => {
+            if (!gqlResult || !gqlResult.fundingEligibility) {
+                throw new Error(`GraphQL fundingEligibility returned no fundingEligibility object`);
+            }
+            return gqlResult && gqlResult.fundingEligibility;
+        });
+    }
+    catch (err) {
```
bluepnume: This should be a `.catch` on the promise, not a try/catch

src/graphql.js
```diff
@@ -0,0 +1,142 @@
...
+        if (status !== 200) {
+            throw new Error(`${ GRAPHQL_URI } returned status ${ status }`);
+        }
+
+        return body.data;
+    });
+}
+
+// call gql with multiple merchant ids to get fundingEligibility for card
+export function getCardEligibility() : ZalgoPromise<CardEligibility> {
```
bluepnume:
```
export function getFundingEligibility<T>() : ZalgoPromise<T> {
````

PR #26, opened 2020-04-24
collects all attributes and use all validate ones in script

weihou:


src/meta.js
```diff
@@ -2,12 +2,11 @@
...
 export function getSDKMeta() : string {
-    return base64encode(JSON.stringify({
-        url:          getScriptUrl(),
-        stageHost:    getStageHost(),
-        apiStageHost: getAPIStageHost()
-    })).replace(/\=+$/, ''); // eslint-disable-line no-useless-escape
+    const attrs = getSDKAttributes();
+
+    attrs.url = getScriptUrl();
+    return base64encode(JSON.stringify(attrs)).replace(/\=+$/, ''); // eslint-disable-line no-useless-escape
```
bluepnume, line 11: Can we filter just `data-` attributes here?

bluepnume: Wait, ignore this comment. `getSDKAttributes` is already doing this.

src/meta.js
```diff
@@ -2,12 +2,11 @@
...
 
 export function getSDKMeta() : string {
-    return base64encode(JSON.stringify({
-        url:          getScriptUrl(),
-        stageHost:    getStageHost(),
-        apiStageHost: getAPIStageHost()
-    })).replace(/\=+$/, ''); // eslint-disable-line no-useless-escape
+    const attrs = getSDKAttributes();
+
+    attrs.url = getScriptUrl();
```
bluepnume, line 10: Can we make the structure `{ url, attrs: { ... } }`

bluepnume:
Looks great. One minor comment & will merge and publish. Thanks Wei!

src/meta.js
weihou: https://github.com/paypal/paypal-sdk-client/blob/master/src/script.js#L34
`getSDKAttributes()` only collects `data-` attributes.

src/meta.js
weihou: do you want the object will be `{
    url:          '',
    attrs:        {
        'data-stage-host':      '',
        'data-api-stage-host':  '',
        'data-merchant-id': ''
    }
}`?

bluepnume: Yep exactly 👍

weihou: fixed. also, had to fix the schema everywhere in the tests

PR #41, opened 2020-07-29
feature/pageType

mnicpt:
Adds the `data-page-type` attribute to the script tag and logs as `pp_placement` key for analytics.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -6,6 +6,8 @@ import { COUNTRY, SDK_SETTINGS, SDK_QUERY_KEYS, INTENT, COMMIT, VAULT, CURRENCY,
 
 import { getHost, getPath, getDefaultStageHost, getDefaultAPIStageHost, getEnv, getDefaultNamespace } from './globals';
 
+const PAGE_TYPES = [ 'home', 'product-list', 'product-detail', 'cart', 'payment' ];
+
 export const CLIENT_ID_ALIAS = {
     sb: 'AZDxjDScFpQtjWTOUtWKbyN_bDt4OgqaF4eYXlewfBP4-8aqX3PiV8e1GWU6liB2CUXlkA59kJXE7M6R'
 };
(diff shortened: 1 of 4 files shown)


src/script.js
```diff
@@ -182,6 +184,13 @@ export function getPartnerAttributionID() : ?string {
     return getSDKAttribute(SDK_SETTINGS.PARTNER_ATTRIBUTION_ID);
 }
 
+export function getPageType() : ?string {
+    const pageType = getSDKAttribute(SDK_SETTINGS.PAGE_TYPE);
+    const validPageType = pageType ? PAGE_TYPES[pageType.toUpperCase()] : false;
+
+    return validPageType ? pageType : '';
```
gregjopa, line 191: I was wondering, do we want to force the pageType to always be uppercase? For example, if page type is passed in as lowercase as 'home' (`'data-page-type': 'home'`), should we log it as 'HOME'? If so we'd need to tweak this. Ex:

```suggestion
    return validPageType ? pageType.toUpperCase() : '';
```

mnicpt: Yes.  If you look at the `PAGE_TYPES` enum, they are all upper case.  So the customer can pass whatever and we convert it to upper case to match the key.

gregjopa: I see that. I guess my question is, should `getPageType()` always return as uppercase? Do we always want them logged as uppercase for consistency? I see in the unit tests that it currently returns as lowercase 'home' if that's what is passed in as the data attribute. So the end user can technically pass in either "home" or "HOME", both and are valid with the ENUM check.

src/script.js
```diff
@@ -182,6 +184,13 @@ export function getPartnerAttributionID() : ?string {
     return getSDKAttribute(SDK_SETTINGS.PARTNER_ATTRIBUTION_ID);
 }
 
+export function getPageType() : ?string {
+    const pageType = getSDKAttribute(SDK_SETTINGS.PAGE_TYPE, '');
+    const validPageType = pageType.length && PAGE_TYPES[pageType.toUpperCase()] !== undefined;
```
bluepnume, line 189: Can we do `values(PAGE_TYPES).indexOf(pageType) !== -1` (`values` can be imported from `belter/src`).

If invalid, can we throw an error to let the merchant know

mnicpt: Updated code and tests.

PR #47, opened 2020-09-29
setup getUserACcessToken function

vishakha94:


bluepnume:
sdk-client should not be aware of zoid or xprops. You can just directly read window.xprops in your child frame.

PR #52, opened 2020-11-17
Update logging to support uid-auto attribute

gregjopa:
I noticed that the initial script uses `data-uid-auto` as the data attribute for storing the uid which is causing the logger to fire a `sdk_non_paypal_domain_script_uid_missing` event. I'm assuming it should be firing a `sdk_non_paypal_domain_script_uid_present` event instead.

Screen Shot 2020-11-16 at 9 41 39 PM

This PR updates the logging logic to work for both `data-uid` and `data-uid-auto`.

This commit introduced the data-uid-auto change: https://github.com/krakenjs/belter/commit/d4fee71912241da03ba039a032b7d22e0ba710e7.

### DIFF
--- a/src/tracking.js
+++ b/src/tracking.js
@@ -24,7 +24,7 @@ export function setupLogger() {
     const logger = getLogger();
 
     sdkInitTime = Date.now();
-    
+
     logger.addPayloadBuilder(() => {
         return {
             referer: window.location.host,
@@ -76,7 +76,7 @@ export function setupLogger() {
         const sdkScript = getSDKScript();
         const loadTime = getResourceLoadTime(sdkScript.src);
         let cache;
-    
+
         if (loadTime === 0) {
             cache = 'sdk_client_cache_hit';
         } else if (typeof loadTime === 'number') {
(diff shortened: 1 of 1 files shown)


mnicpt (approved):
Good catch!

bluepnume:
No, this is expected. `data-uid-auto` means "`data-uid` was not present, so we automatically added a uid -- but there's no guarantee it will match the uid from the parent window". We need that guarantee.

bluepnume:
So any app showing `sdk_non_paypal_domain_script_uid_missing` is likely to be an app which does not have the latest version of `@paypal/sdk-client`. These apps would need to upgrade and roll out. Then they should start seeing `data-uid` being correctly set in the server-side renderer into the script tag.

gregjopa:
Thanks @bluepnume. I figured I was missing something. 

> data-uid-auto means "data-uid was not present, so we automatically added a uid -- but there's no guarantee it will match the uid from the parent window". We need that guarantee.

If I understand correctly, the parent window will always have `data-uid-auto` added to the script and will always fire a `sdk_non_paypal_domain_script_uid_missing` event. What we care about measuring is the apps powering the iframes and checking to see if they inherit the uid from the parent correctly or not. Does that sound right?

I'll go ahead and close this one out.

bluepnume:
> What we care about measuring is the apps powering the iframes and checking to see if they inherit the uid from the parent correctly or not. Does that sound right?

Exactly yeah 👍

PR #56, opened 2020-12-14
Update belter to support data-uid

gregjopa:
Support for `data-uid` was added to belter@1.0.136 so we need sdk-client to be using at least that. 

https://github.com/krakenjs/belter/compare/v1.0.135...v1.0.136

I ran into this problem when updating a private server-side app to use the latest sdk-client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
   "dependencies": {
     "@paypal/sdk-constants": "^1.0.61",
     "beaver-logger": "^4.0.0",
-    "belter": "^1.0.16",
+    "belter": "^1.0.136",
     "bowser": "^2.0.0",
     "jsx-pragmatic": "^2",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -51,7 +51,7 @@
   "dependencies": {
     "@paypal/sdk-constants": "^1.0.61",
     "beaver-logger": "^4.0.0",
-    "belter": "^1.0.16",
+    "belter": "^1.0.136",
```
gregjopa, line 54: The latest version of belter is `1.0.146`. Should we use that instead? Happy to update it.

I went with 1.0.136 since that's when `ATTRIBUTES.UID` was added.

gregjopa: I changed my mind and went with the latest belter `1.0.146`. My thinking is it's a util library so it's always best to have the latest.

PR #58, opened 2021-01-25
Allow local testing

mnicpt:
This allows for local testing of changes on native devices.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "setup": "npm install && npm run flow-typed",
     "test": "npm run lint && npm run flow-typed && npm run flow && npm run jest && npm run karma",
     "webpack": "babel-node --plugins=transform-es2015-modules-commonjs ./node_modules/.bin/webpack --progress",
-    "jest": "jest test/server --env=node --no-cache --collectCoverageFrom='server/' --coverage --verbose --runInBand --silent=false",
+    "jest": "NODE_ENV=local jest test/server --env=node --no-cache --collectCoverageFrom='server/' --coverage --verbose --runInBand --silent=false",
     "prepublishOnly": "npm run babel",
     "postpublish": "rm -rf ./server && git checkout ./server"
   },
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "setup": "npm install && npm run flow-typed",
     "test": "npm run lint && npm run flow-typed && npm run flow && npm run jest && npm run karma",
     "webpack": "babel-node --plugins=transform-es2015-modules-commonjs ./node_modules/.bin/webpack --progress",
-    "jest": "NODE_ENV=local jest test/server --env=node --no-cache --collectCoverageFrom='server/' --coverage --verbose --runInBand --silent=false",
+    "jest": "jest test/server --env=node --no-cache --collectCoverageFrom='server/' --coverage --verbose --runInBand --silent=false",
     "prepublishOnly": "npm run babel",
     "postpublish": "rm -rf ./server && git checkout ./server"
   },
--- a/server/constants.js
+++ b/server/constants.js
@@ -1,9 +1,12 @@
 /* @flow */
 
 export const HOST = {
-    LOCALHOST:     'localhost.paypal.com',
-    PAYPAL:        '.paypal.com',
-    PAYPALOBJECTS: 'www.paypalobjects.com'
+    LOCALHOST:      'localhost.paypal.com',
+    PAYPAL:         '.paypal.com',
+    PAYPALOBJECTS:  'www.paypalobjects.com',
+    LOCALTUNNEL:    'paypal.loca.it',
+    LOCALHOST_8000: 'localhost:8000',
+    LOCALHOST_8443: 'localhost:8443'
 };
 
 export const PROTOCOL = {
--- a/server/constants.js
+++ b/server/constants.js
@@ -4,7 +4,7 @@ export const HOST = {
     LOCALHOST:      'localhost.paypal.com',
     PAYPAL:         '.paypal.com',
     PAYPALOBJECTS:  'www.paypalobjects.com',
-    LOCALTUNNEL:    'paypal.loca.it',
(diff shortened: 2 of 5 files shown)


server/meta.jsx
```diff
@@ -74,6 +74,11 @@ function isSDKUrl(hostname : string) : boolean {
     return false;
 }
 
+function isLocalUrl(hostname : string) : boolean {
+    const localUrls = ['localhost', 'ngrok', 'loca'];
+    return localUrls.some(url => hostname.includes(url));
```
bluepnume, line 79: Can't do `.includes()` since it would match something like `localhost.attacksite.com`. Would need to define an array of full domains we want to match.

src/domains.js
```diff
@@ -105,6 +105,10 @@ export function getOrderAPIUrl() : string {
 }
 
 export function getPayPalDomainRegex() : RegExp {
+    if (__ENV__ === ENV.LOCAL) {
+        // eslint-disable-next-line security/detect-non-literal-regexp
+        return new RegExp(`${ getProtocol() }://${ getHost() }`);
```
bluepnume, line 110: There might be problems here since `getHost()` contains special characters like `.` which will be turned into a regex. Can we just add localhost to the existing regex?

mnicpt: I have us supporting domains with ngrok, localhost or loca.  Will add all.

mnicpt: Maybe keep the if condition so that they aren't proxied in production to some malicious person

server/meta.jsx
mnicpt: @bluepnume Provided definitive hosts.

server/constants.js
```diff
@@ -1,9 +1,12 @@
 /* @flow */
 
 export const HOST = {
-    LOCALHOST:     'localhost.paypal.com',
-    PAYPAL:        '.paypal.com',
-    PAYPALOBJECTS: 'www.paypalobjects.com'
+    LOCALHOST:      'localhost.paypal.com',
+    PAYPAL:         '.paypal.com',
+    PAYPALOBJECTS:  'www.paypalobjects.com',
+    LOCALTUNNEL:    'paypal.loca.lt',
```
gregjopa, line 7: Can you share more details about `paypal.loca.lt`? Could we technically be attacked if someone go a hold of this domain on [localtunnel](https://theboroer.github.io/localtunnel-www/)?

mnicpt: @gregjopa For sure!  These local url's are only evaluated when `env=local`.  This is our unified way of being able to test SPB changes locally on native devices.

server/meta.jsx
```diff
@@ -74,8 +74,13 @@ function isSDKUrl(hostname : string) : boolean {
     return false;
 }
 
+function isLocalUrl(host : string) : boolean {
+    const localUrls = [HOST.LOCALHOST_8000, HOST.LOCALHOST_8443, HOST.LOCALTUNNEL];
+    return localUrls.some(url => host === url);
```
gregjopa, line 79: To be safe, could we also add an environment check here?
```suggestion
    return __ENV__ === ENV.LOCAL && localUrls.some(url => host === url);
```

server/constants.js
gregjopa: @mnicpt thanks for explaining. So as long as we have the `__ENV__ === ENV.LOCAL` check for these urls, we prevent any security vulnerabilities. That makes sense.

server/meta.jsx
mnicpt: @gregjopa We could but this cannot occur in production.  The config.json will cause this to fail due to the CSP policy.  But don't mind adding for extra security.

mnicpt: @gregjopa __ENV__ is not available on the server.  I think we're good here.  @bluepnume ?

bluepnume: Can we check `process.env.NODE_ENV`? I think this is set by most paypal apps. May need to add a log to check if the correct value is `localhost` or `development` or something

test/server/meta.test.js
```diff
@@ -69,8 +69,27 @@ test('should fail to construct a script url with checkout.js on localhost withou
...
+        unpackSDKMeta(Buffer.from(JSON.stringify({
+            url: sdkUrl
+        })).toString('base64'));
+    } catch (err) {
+        error = err;
+    }
+
     if (!error) {
-        throw new Error(`Expected error to be thrown`);
+        throw new Error(`Should construct script with supported local urls: (localhost, ngrok, loca)`);
```
gregjopa, line 92: I think we removed `ngrok` as an option, right? Can we remove it here as well?
```suggestion
        throw new Error(`Should construct script with supported local urls: (localhost, loca)`);
```

mnicpt: Yes. Thank you for the catch.  Don't want to confuse the requirements.

package.json
```diff
@@ -21,7 +21,7 @@
     "setup": "npm install && npm run flow-typed",
     "test": "npm run lint && npm run flow-typed && npm run flow && npm run jest && npm run karma",
     "webpack": "babel-node --plugins=transform-es2015-modules-commonjs ./node_modules/.bin/webpack --progress",
-    "jest": "jest test/server --env=node --no-cache --collectCoverageFrom='server/' --coverage --verbose --runInBand --silent=false",
+    "jest": "NODE_ENV=local jest test/server --env=node --no-cache --collectCoverageFrom='server/' --coverage --verbose --runInBand --silent=false",
```
gregjopa, line 24: I think jest sets NODE_ENV to "test" by default. Can we stick with that and mock out this value for the specific test that is expecting `NODE_ENV === "test"` to be true?

mnicpt: @gregjopa That's correct.  It is set to test.  Not sure how this matters, but will see about overriding for that test.

bluepnume (approved):
LGTM! Can you check the tests and feel free to merge after. 👍

gregjopa (approved):

PR #59, opened 2021-02-05
Added sandbox and test client-id aliases

mstuart:
This is pretty minor, but the Dev Docs team thought it'd be a good idea. I don't see why not. "sb" might not be obvious to people right away. 

Added sandbox and test client-id aliases, then we'll update the https://developer.paypal.com/demo/checkout/#/pattern/client tool and other places it appears.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -13,8 +13,12 @@ export const SUPPORTED_BROWSERS = {
     vivaldi:        '1.91'
 };
 
+const SANDBOX_ALIAS = 'AZDxjDScFpQtjWTOUtWKbyN_bDt4OgqaF4eYXlewfBP4-8aqX3PiV8e1GWU6liB2CUXlkA59kJXE7M6R';
+
 export const CLIENT_ID_ALIAS = {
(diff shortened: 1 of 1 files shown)


src/config.js
```diff
@@ -14,7 +14,9 @@ export const SUPPORTED_BROWSERS = {
 };
 
 export const CLIENT_ID_ALIAS = {
-    sb: 'AZDxjDScFpQtjWTOUtWKbyN_bDt4OgqaF4eYXlewfBP4-8aqX3PiV8e1GWU6liB2CUXlkA59kJXE7M6R'
+    sandbox:    'AZDxjDScFpQtjWTOUtWKbyN_bDt4OgqaF4eYXlewfBP4-8aqX3PiV8e1GWU6liB2CUXlkA59kJXE7M6R',
```
gregjopa, line 17: Could you create a variable for this client-id and reuse it for sb, sandbox, and test? 

I'm thinking about bundle size. I don't think babel would be smart enough to transpile this down to one string variable in the current format.

mstuart: good catch. done! 👍

gregjopa (approved):
Looks great!

bluepnume:
FYI, will need to add the same to clientsdknodeweb code.

bluepnume (approved):

mnicpt (approved):
Phenomonal!

PR #65, opened 2021-04-06
support for sandbox v2 added

gabrielo91:
# Description

- Now it's possible to switch between two sandbox environments, this PR adds support for such a scenario
- New tests added for  `getPayPalDomain` and  `getPayPalAPIDomain`  functions
- Coverage in `domain.js` increased up to 60%

## Results of new testing setup for client

image

## Collaboration:
Thanks to @bluepnume for [this update](https://github.com/krakenjs/grumbler-scripts/commit/2f9d47fcf602bad7f12a19e1aef3fb398e2375df) that allow us to avoid struggling with globals in test environment


mnicpt (approved):
Looks great!

src/domains.js
```diff
@@ -24,6 +24,10 @@ export function getPayPalDomain() : string {
     if (__ENV__ === ENV.SANDBOX) {
         return `${ getProtocol() }://www.sandbox.paypal.com`;
     }
+    
+    if (__ENV__ === ENV.SANDBOX_V2) {
+        return `${ getProtocol() }://www.v2.sandbox.paypal.com`;
```
mnicpt, line 29: This is supposed to be `www-v2.sandbox.paypal.com`

mnicpt (changes requested):
Change www url.

src/domains.js
mnicpt: I actually wonder if it's incorrect in the requirements.  Will check.

gabrielo91: I thought it was an error, since`www-v2` didn´t seem like an ivalid value, but just checked in Jira and you´re right. 

Changed!

mnicpt: Verified with contact and he said it needs to be with the dash.

package.json
```diff
@@ -49,7 +49,7 @@
   "license": "Apache-2.0",
   "readmeFilename": "README.md",
   "dependencies": {
-    "@paypal/sdk-constants": "^1.0.94",
+    "@paypal/sdk-constants": "^1.0.97",
```
mnicpt, line 52: This is fine, but FYI, you don't need to change it since it has the `^` and will automatically update when you reinstall node_modules for minor versions.

gabrielo91: oh! I forgot that, thanks!

mnicpt (approved):

mnicpt (approved):
LGTM!  Looks great!

mnicpt (changes requested):
Can we add some test?

src/domains.js
```diff
@@ -24,6 +24,10 @@ export function getPayPalDomain() : string {
     if (__ENV__ === ENV.SANDBOX) {
         return `${ getProtocol() }://www.sandbox.paypal.com`;
     }
+    
+    if (__ENV__ === ENV.SANDBOX_V2) {
```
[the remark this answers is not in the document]
bluepnume: If you want to avoid modifying `ENV` could always just create a new temporary global flag like `__IS_SANDBOX_V2__`. Makes it easy to find/replace it later when we don't need it any more.

gabrielo91:
> Can we add some tests?
@mnicpt 

At a first glance, there are some problems with, at least due to the way how test are wroten now, it turns out that [WebPack](https://github.com/paypal/paypal-sdk-client/blob/master/webpack.config.js) imports thee environment variables from [here](https://github.com/paypal/paypal-sdk-client/blob/master/test/globals.js) and it does in build time and they can't be overwritten or switched in test time in order to test the `getPayPalDomain` function. @gregjopa  and I spent some minutes exploring another alternative and, maybe it's possible to solve it with a new and isolated `jest` testing setup, let me try it for a while and I'll let u know.

package.json
```diff
@@ -13,6 +13,7 @@
     "flow-typed": "rm -rf ./flow-typed && flow-typed install",
     "karma": "cross-env NODE_ENV=test babel-node --plugins=transform-es2015-modules-commonjs ./node_modules/.bin/karma start",
     "lint": "eslint src/ server/ test/ *.js",
+    "lint:fix": "eslint --fix src/ server/ test/ *.js",
```
mnicpt, line 16: I've seen that this can be dangerous.  Can we remove this?

mnicpt: We are caught up on linting for this repo so no reason we can't fix individually.

test/jest/domain.test.js
```diff
@@ -1,99 +1,100 @@
...
-  insertMockSDKScript
-} from "../../src";
+    getPayPalDomain,
+    getPayPalAPIDomain,
+    insertMockSDKScript
+} from '../../src';
 
 beforeEach(() => {
-  __ENV__ = "test";
+    __ENV__ = 'test';
```
mnicpt, line 9: Has indentation changed?  Looks like 4 spaces now.

package.json
bluepnume: This is fine, you have to run this manually. 👍 same as `npm run lint -- --fix`

bluepnume: (FWIW can also set up vscode to auto-fix on save, super useful)

gabrielo91: And it is! I can, however my default configuration seems not to be the same required here, so my auto formatting is not suitable. I use `esbenp.prettier-vscode` ... any suggestion about how to configure it properly??

I'll remove that line btw

test/jest/domain.test.js
```diff
@@ -0,0 +1,128 @@
...
+});
+
+describe(`getPayPalDomain function test cases`, () => {
+    it('should successfully get the production paypal domain', () => {
+        __ENV__ = 'production';
+        const expectedPayPalDomain = 'https://www.paypal.com';
+
+        if (getPayPalDomain() !== expectedPayPalDomain) {
+            throw new Error(`Expected paypal domain to be ${ expectedPayPalDomain }, got ${ getPayPalDomain() }`);
```
gregjopa, line 19: `expect` should be fine to use for Jest assertions. I think we just need to tell eslint about the Jest globals.

gregjopa: Great job on these tests @gabrielo91! 💯

gabrielo91: Yes, I can avoid eslint throw errors, however, the build still fails because of @flow  (I guess, I can check that if we definitely want to use `expect`

image

gabrielo91: I found a way, but I don't know how bad is it, since is not clear to me what `flow`  does exactly... Said that... We  can add 
`.*\.test\.js` on .flowconfig and the build works... however, and again.. not sure if is the right way... another one according to what I read, is to add an extra plugin for babel, but such option seems less attractive

gabrielo91: I have another commit with that approach (jest-expect) and the build working, so we can switch between them and chose the right one ... regarding the coverage, I think that more files are had in the account now, so we need to adjust that (It's not because test coverage was reduced)

gregjopa: I wonder if we can use this babel plugin to get jest working with flow: 
- https://babeljs.io/docs/en/babel-plugin-transform-flow-strip-types.html. 
- https://www.grzegorowski.com/jest-tests-flow-type

test/jest/domain.test.js
```diff
@@ -0,0 +1,128 @@
...
+        }
+    });
+
+    it('should successfully get the stage paypal domain', () => {
```
gregjopa, line 23: This is minor but what do you think about this grouping for the tests? It consolidates the tests more and uses `test` instead of `it` to stay consistent with the server-side jest tests.

```js
describe('getPayPalDomain()', () => {
    test('should get the expected paypal domain for each environment', () => {
        __ENV__ = 'production';
        expect(getPayPalDomain()).toBe('https://www.paypal.com');

        __ENV__ = 'sandbox';
        expect(getPayPalDomain()).toBe('https://www.sandbox.paypal.com');

        __ENV__ = 'stage';
        expect(getPayPalDomain()).toBe('https://msmaster.qa.paypal.com');
    });

    test('should get the expected paypal domain when using query param env=sandbox_v2', () => {
        insertMockSDKScript({
            query: {
                'client-id': 'sb',
                'env':       'sandbox_v2'
            }
        });

        __ENV__ = 'sandbox';
        expect(getPayPalDomain()).toBe('https://www-v2.sandbox.paypal.com');
    });

    test('should throw an error when the environment is not found', () => {
        __ENV__ = 'invalid_environment';
        expect(getPayPalDomain()).toThrow(`Can not get paypal domain for env: ${ __ENV__ }`);
    });

})

```

bluepnume (approved):

bluepnume:
Looks great to me!

test/client/config.js
```diff
@@ -1,25 +1,133 @@
...
+    });
+
+    it('should successfully get the sandbox v2 paypal domain', () => {
+        const expectedPayPalDomain = 'https://www-v2.sandbox.paypal.com';
+        window.__ENV__ = 'sandbox';
+
+        insertMockSDKScript({
+            query: {
+                'client-id': 'sb',
+                'env':         'sandbox_v2'
```
gregjopa, line 50: Nit: this spacing seems a little off. I think `sb` and `sandbox_v2` should align. Maybe its the github ui

gregjopa (approved):
Great work @gabrielo91! Thanks for investing the time to get good unit tests in with this change 💯

test/client/config.js
gabrielo91: I was my `lint --fix ` fault, fixed

mnicpt:
Great job, @gabrielo91 !  Way to persevere :)

PR #66, opened 2021-04-09
Jest expect testing approach

gabrielo91:


bluepnume:
Happy for us to merge this, and would be good to explore Jest (although would want to make sure it plays nicely with how we use webpack, so our tests are as close to production builds as possible).

That said, it is actually totally possible to change globals with the current testing setup. Here's an example: https://github.com/paypal/paypal-sdk-client/commit/4d7590721267a036f55de29dbefe018583029f61

gregjopa:
Great point @bluepnume! We didn't realize this was possible with karma and assumed they were build time settings. We will go with that for this sandbox-v2 feature. 

Getting jest to play nice with our existing webpack/flow grumbler-scripts strategy would be great but doesn't have to be tied to this sandbox-v2 work.

PR #67, opened 2021-04-20
Update getCSPNonce to always return a string

gregjopa:
This PR updates `getCSPNonce` to return an empty string by default. This will make it easier enforce the csp nonce values in paypal-checkout-components by making them required strings.


mstuart:
nice! :)

mstuart (approved):

PR #68, opened 2021-04-22
[DTCHKINT-244] Check for single country match from locale lang

amyegan:
Sometimes when pulling locale from browser settings, we only receive a language without a country code. In cases where there is only one possible country match, we should be able to safely infer the correct country from the language provided.

- Check that there is only one possible country match for the provided language and use it for locale
- Add tests for `getLocale` to cover new code change and some existing cases

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -224,20 +224,25 @@ export function getLocale() : LocaleType {
             return { country, lang };
         } else {
             // We will infer country from language if there is only one possible country match
-            const possibleCountries = Object.keys(COUNTRY_LANGS).filter((possibleCountry) => {
+            let possibleLocales = [];
+            for(let possibleCountry of Object.keys(COUNTRY_LANGS)) {
                 const languages = COUNTRY_LANGS[possibleCountry];
 
+                console.log('possible country', possibleCountry);
+                // console.log('languages', languages);
+
                 if (languages.length && languages.some(language => language === lang)) {
-                    return true;
+                    const language = languages.find(x => x === lang);
+                    possibleLocales.push({country: possibleCountry, lang: language});
                 }
 
-                return false;
-            });
+                if (possibleLocales.length > 1) {
+                    break;
+                }
+            }
 
-            const languageKey = Object.keys(LANG).filter(possibleLanguage => LANG[possibleLanguage] === lang)[0];
-            const language = LANG[languageKey];
-            if (possibleCountries.length === 1 && COUNTRY[possibleCountries[0]] && language) {
-                return { country: possibleCountries[0], lang: language };
+            if (possibleLocales.length === 1) {
+                return possibleLocales[0];
             }
         }
     }
--- a/src/script.js
+++ b/src/script.js
@@ -223,23 +223,19 @@ export function getLocale() : LocaleType {
             // $FlowFixMe
             return { country, lang };
         } else {
-            // We will infer country from language if there is only one possible country match
-            const possibleLocales = [];
-            for (const possibleCountry of Object.keys(COUNTRY_LANGS)) {
-                const languages = COUNTRY_LANGS[possibleCountry];
-                const language = languages.find(x => x === lang);
-
-                if (language) {
-                    possibleLocales.push({ country: possibleCountry, lang: language });
-                }
-
-                if (possibleLocales.length > 1) {
-                    break;
-                }
+            // We infer country from language if there is only one possible country match
+            const possibleCountries = Object.keys(COUNTRY_LANGS).filter(x => COUNTRY_LANGS[x].some(y => y === lang));
+            
+            if (possibleCountries.length !== 1) {
+                break;
             }
 
-            if (possibleLocales.length === 1 && possibleLocales[0]) {
-                return possibleLocales[0];
+            const possibleCountry = possibleCountries[0];
+            const languages = COUNTRY_LANGS[possibleCountry];
+            const language = languages.find(x => x === lang);
+
+            if (language) {
+                return { country: possibleCountry, lang: language };
             }
         }
     }
--- a/src/script.js
+++ b/src/script.js
@@ -224,16 +224,13 @@ export function getLocale() : LocaleType {
             return { country, lang };
         } else {
             // We will infer country from language if there is only one possible country match
-            let possibleLocales = [];
-            for(let possibleCountry of Object.keys(COUNTRY_LANGS)) {
+            const possibleLocales = [];
+            for (const possibleCountry of Object.keys(COUNTRY_LANGS)) {
                 const languages = COUNTRY_LANGS[possibleCountry];
+                const language = languages.find(x => x === lang);
 
-                console.log('possible country', possibleCountry);
-                // console.log('languages', languages);
-
-                if (languages.length && languages.some(language => language === lang)) {
-                    const language = languages.find(x => x === lang);
-                    possibleLocales.push({country: possibleCountry, lang: language});
+                if (language) {
+                    possibleLocales.push({ country: possibleCountry, lang: language });
                 }
 
                 if (possibleLocales.length > 1) {
@@ -241,7 +238,7 @@ export function getLocale() : LocaleType {
                 }
             }
 
-            if (possibleLocales.length === 1) {
+            if (possibleLocales.length === 1 && possibleLocales[0]) {
                 return possibleLocales[0];
             }
         }
--- a/src/script.js
+++ b/src/script.js
@@ -224,27 +224,21 @@ export function getLocale() : LocaleType {
             return { country, lang };
         } else {
             // We will infer country from language if there is only one possible country match
-            // $FlowFixMe
-            const possibleCountries = [];
-            // $FlowFixMe
-            for (const possibleCountry in COUNTRY_LANGS) {
-                if (COUNTRY_LANGS.hasOwnProperty(possibleCountry)) {
-                    if (possibleCountries.length > 1) {
-                        break;
-                    }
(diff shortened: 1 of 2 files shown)


mnicpt (approved):
Great job, Amy! Can you fix all the flow errors and remove all of the `// $FlowFixMe`?

test/client/script.js
```diff
@@ -340,4 +340,77 @@ describe(`script cases`, () => {
...
+        if (expectedLocale !== receivedLocal) {
+            throw new Error(`Expected client id to be ${ expectedLocale }, got ${ receivedLocal } from ${ url }`);
+        }
+    });
+
+    it('should successfully get locale from browser settings', () => {
+        const expectedLocale = 'en_US';
+
+        // $FlowFixMe
+        Object.defineProperty(navigator, 'languages', {
```
mnicpt, line 364: Do you need to do this?

test/client/script.js
```diff
@@ -340,4 +340,77 @@ describe(`script cases`, () => {
...
+        if (expectedLocale !== receivedLocale) {
+            throw new Error(`Expected client id to be ${ expectedLocale }, got ${ receivedLocale }`);
+        }
+    });
+
+    it('should return default locale if none detected', () => {
+        const expectedLocale = 'en_US';
+
+        // $FlowFixMe
+        Object.defineProperty(navigator, 'languages', {
```
mnicpt, line 398: Same.

test/client/script.js
```diff
@@ -340,4 +340,77 @@ describe(`script cases`, () => {
...
+            get:          () => [],
+            configurable: true
+        });
+
+        // $FlowFixMe
+        Object.defineProperty(navigator, 'language', {
```
mnicpt, line 404: Same.

mnicpt:
@amyegan Sorry for the extra comments.  Didn't see the tests initially.

test/client/script.js
amyegan: That didn't work since languages is read-only. I had to use `Object.defineProperty` and set `configurable:true` to get it set the languages in the test. I have modified it to remove the $FlowFixMe and make it maybe a little easier to read

test/client/script.js
```diff
@@ -340,4 +340,76 @@ describe(`script cases`, () => {
...
+        if (expectedLocale !== receivedLocal) {
+            throw new Error(`Expected client id to be ${ expectedLocale }, got ${ receivedLocal } from ${ url }`);
+        }
+    });
+
+    it('should successfully get locale from browser settings', () => {
+        const expectedLocale = 'en_US';
+        const defineProp = Object.defineProperty;
+
+        defineProp(navigator, 'languages', {
```
gregjopa, line 364: Thanks for adding tests @amyegan! 💯 

I've had to do the same with making globals writable in jest. One suggestion is to do this in the `beforeEach` like so:

```js
beforeEach(() => {
    Object.defineProperty(window.navigator, 'languages', { value: [], writable: true });
});

// and then in the test do the following
window.navigator.languages = ['en_US'];
```

src/script.js
```diff
@@ -222,6 +222,24 @@ export function getLocale() : LocaleType {
...
+                const languages = COUNTRY_LANGS[possibleCountry];
+
+                if (languages.length && languages.some(language => language === lang)) {
+                    return true;
+                }
+
+                return false;
+            });
+
+            const inferredCountry = COUNTRY[possibleCountries[0]];
```
bluepnume, line 237: Can we do this lookup after checking `possibleCountries.length === 1`? Otherwise we're doing `COUNTRY[undefined]` which will technically work but feels error prone.

src/script.js
```diff
@@ -222,6 +222,24 @@ export function getLocale() : LocaleType {
...
+            const languageKey = Object.keys(LANG).filter(possibleLanguage => LANG[possibleLanguage] === lang)[0];
```
bluepnume, line 238: Would it be clearer to just do `if (values(LANG).indexOf(lang) !== -1)` to check if it's a valid language?

amyegan: I didn't know that function was available, but that's very helpful! I was trying to find a way to avoid using `$FlowFixMe` or `$FlowIgnore` and I'm not sure how else I can make flow happy in this case

test/client/script.js
```diff
@@ -3,9 +3,17 @@
 import { base64encode } from 'belter/src';
 
 import { getClientID, getIntent, getCurrency, getVault, getCommit, getClientToken, getPartnerAttributionID,
-    getMerchantID, getClientAccessToken, getSDKIntegrationSource, insertMockSDKScript, getPageType } from '../../src';
+    getMerchantID, getClientAccessToken, getSDKIntegrationSource, insertMockSDKScript, getPageType, getLocale } from '../../src';
 
 describe(`script cases`, () => {
+    beforeEach(() => {
+        /* eslint-disable compat/compat */
```
gregjopa, line 10: @amyegan what do you think about using `/* eslint-disable compat/compat */` for the whole test file? My initial thought is none of the test files should care about the `compat` rules.

src/script.js
gregjopa: +1 for using the `values()` helper function from belter assuming it plays nice with Flow.

mnicpt: That's so belter!

amyegan: I ended up changing this whole else clause so I no longer need to do a bunch of weird things to make Flow happy

src/script.js
```diff
@@ -222,6 +222,25 @@ export function getLocale() : LocaleType {
...
             return { country, lang };
+        } else {
+            // We will infer country from language if there is only one possible country match
+            const possibleLocales = [];
+            for (const possibleCountry of Object.keys(COUNTRY_LANGS)) {
+                const languages = COUNTRY_LANGS[possibleCountry];
+                const language = languages.find(x => x === lang);
+
+                if (language) {
+                    possibleLocales.push({ country: possibleCountry, lang: language });
```
bluepnume, line 233: Could we simplify this by just doing a `return` here, rather than building an array and breaking out of the loop?

mstuart:
lgtm! 👍  @gregjopa @mnicpt @bluepnume, can you take another look?

gregjopa (approved):
Great work @amyegan! The unit test coverage looks solid. Awesome job! 💯 🥇 👍

test/client/script.js
```diff
@@ -340,4 +347,56 @@ describe(`script cases`, () => {
...
+
+        const localeObject = getLocale();
+        const receivedLocale = `${ localeObject.lang }_${ localeObject.country }`;
+
+        if (expectedLocale !== receivedLocale) {
+            throw new Error(`Expected client id to be ${ expectedLocale }, got ${ receivedLocale }`);
+        }
+    });
 });
+/* eslint-enable compat/compat */
```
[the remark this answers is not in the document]
amyegan: I didn't think we should need that either, but I get lint errors without it

src/script.js
```diff
@@ -222,6 +222,21 @@ export function getLocale() : LocaleType {
...
+        } else {
+            // We infer country from language if there is only one possible country match
+            const possibleCountries = Object.keys(COUNTRY_LANGS).filter(x => COUNTRY_LANGS[x].some(y => y === lang));
+            
+            if (possibleCountries.length !== 1) {
+                break;
+            }
+
+            const possibleCountry = possibleCountries[0];
+            const languages = COUNTRY_LANGS[possibleCountry];
```
mnicpt, line 234: @amyegan Do we have to determine language here since we are already filtering by `lang` to get possible country?  Wouldn't we just use `lang`?

bluepnume (approved):
LGTM!

src/script.js
bluepnume: Yeah this is a fair point

test/client/script.js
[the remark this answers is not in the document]
bluepnume: Recommend we disable the specific line rather than the entire file.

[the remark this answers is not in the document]
gregjopa: @bluepnume I originally told @amyegan to make this change for the whole file. I was thinking that `compat/compat` linting feature shouldn't be used for test files and test files should be able to leverage the latest JS features. Let's change this back to be line specific and we can discuss offline. Sorry for the back and forth.

test/client/script.js
```diff
@@ -340,4 +347,56 @@ describe(`script cases`, () => {
...
+
+    it('should successfully get locale from browser settings', () => {
+        const expectedLocale = 'fr_FR';
+        window.navigator.languages = [ expectedLocale ];
+
+        const localeObject = getLocale();
+        const receivedLocale = `${ localeObject.lang }_${ localeObject.country }`;
+
+        if (expectedLocale !== receivedLocale) {
+            throw new Error(`Expected client id to be ${ expectedLocale }, got ${ receivedLocale }`);
```
mnicpt, line 375: Same

test/client/script.js
```diff
@@ -340,4 +347,56 @@ describe(`script cases`, () => {
...
+
+    it('should infer locale country from language', () => {
+        const expectedLocale = 'ja_JP';
+        window.navigator.languages = [ 'ja' ];
+
+        const localeObject = getLocale();
+        const receivedLocale = `${ localeObject.lang }_${ localeObject.country }`;
+
+        if (expectedLocale !== receivedLocale) {
+            throw new Error(`Expected client id to be ${ expectedLocale }, got ${ receivedLocale }`);
```
mnicpt, line 387: same

test/client/script.js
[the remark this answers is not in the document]
amyegan: No problem. I switched it to only disable specific lines

PR #70, opened 2021-04-27
Add support for using v4 from qa cdn

gregjopa:
This PR adds support for loading checkout.js from the QA CDN.

For example, a test page should be able to leverage the QA CDN like so:

```html
<script src="https://uideploy--staticcontent--7482d416a81b5--ghe.preview.dev.paypalinc.com/api/checkout.js" data-paypal-checkout></script>
```

This PR adds support for trusting the `.paypalinc.com` domain when validating the url stored in sdkMeta that's used for second render.

### DIFF
--- a/server/constants.js
+++ b/server/constants.js
@@ -1,12 +1,13 @@
 /* @flow */
 
 export const HOST = {
-    LOCALHOST:      'localhost.paypal.com',
-    PAYPAL:         '.paypal.com',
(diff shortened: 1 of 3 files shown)


bluepnume (approved):

bluepnume:
Looks good!

mstuart:
awesome work!!! :D :D

PR #71, opened 2021-05-21
Add csp_nonce as an allowed attr for sdk meta

alan-bendjuya:
This PR add support for data-csp-nonce.
The property was missing in the allowed attributes when the sdk meta is encoded. As a result the script tag is created without the value that needs to be passed on.


src/meta.js
```diff
@@ -12,7 +12,8 @@ const ALLOWED_ATTRS = [
     SDK_SETTINGS.ENABLE_3DS,
     SDK_SETTINGS.SDK_INTEGRATION_SOURCE,
     SDK_SETTINGS.CLIENT_METADATA_ID,
-    ATTRIBUTES.UID
+    ATTRIBUTES.UID,
+    SDK_SETTINGS.CSP_NONCE
```
gregjopa, line 16: I was debugging this with @alan-bendjuya and learned that on second render the `data-csp-nonce` value was missing from the sdk script and it's needed so SPB can pass the nonce to the overlay component on button click.

gregjopa (approved):
Looks great! 💯

PR #76, opened 2021-06-11
Move domain logic to the server-side

gregjopa:
This PR updates `getPayPalDomain()` and `getPayPalAPIDomain()` to be global build time variables. This logic moves from the client-side to the server-side for better geo support.


bluepnume (approved):
Awesome

PR #79, opened 2021-06-25
support for special country locales added

gabrielo91:
## Description

- Special Language variants are now supported and are handled as new entire languages. Such strings contain no only the specific country language, but  the country as well, a detailed explanation [here](https://github.com/paypal/paypal-sdk-constants/pull/55) 

## ScreenShots
image

## Related PR's

- https://github.com/paypal/paypal-sdk-constants/pull/55
- https://github.com/paypal/paypal-identity-components/pull/14
- https://github.com/paypal/paypal-checkout-components/pull/1637

## Depends on:

- https://github.paypal.com/Checkout-R/clientsdknodeweb/pull/304

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,6 +1,6 @@
 /* @flow */
 import type { LocaleType } from '@paypal/sdk-constants/src/';
-import { COUNTRY_LANGS } from '@paypal/sdk-constants/src';
+import { COUNTRY_LANGS, LANG } from '@paypal/sdk-constants/src';
 
 
 export function getComputedLocales(locale : string) : LocaleType {
@@ -8,8 +8,8 @@ export function getComputedLocales(locale : string) : LocaleType {
     // $FlowFixMe
     const countryLangs = COUNTRY_LANGS[country];
 
-    if (countryLangs.includes('zh_Hant')) {
-        lang = 'zh_Hant';
+    if (countryLangs.includes(LANG.ZH_HANT)) {
+        lang = LANG.ZH_HANT;
     }
 
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -3,6 +3,7 @@ import type { LocaleType } from '@paypal/sdk-constants/src';
 
 const SPECIAL_COUNTRY_LANG_VARIANTS = {
     'HK': 'zh_Hant'
+    'TW': 'zh_Hant'
 };
 
 export function getComputedLocales(locale : string) : LocaleType {
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,16 +1,17 @@
 /* @flow */
 import type { LocaleType } from '@paypal/sdk-constants/src/';
-import { COUNTRY_LANGS, LANG } from '@paypal/sdk-constants/src';
+import { COUNTRY_LANGS, LANG, COUNTRY } from '@paypal/sdk-constants/src';
 
 export function getComputedLocales(locale : string) : LocaleType {
-    let [ lang, country ] = locale.split('_');
-    // $FlowFixMe
+    const [ _lang, _country ] = locale.split('_');
+    
+    let lang = LANG[_lang.toUpperCase()];
+    const country = COUNTRY[_country];
     const countryLangs = COUNTRY_LANGS[country];
 
     if (countryLangs.includes(LANG.ZH_HANT)) {
         lang = LANG.ZH_HANT;
     }
 
-    // $FlowFixMe
     return { lang, country };
 }
--- a/src/utils.js
+++ b/src/utils.js
@@ -2,7 +2,7 @@
 import type { LocaleType } from '@paypal/sdk-constants/src';
 
 const SPECIAL_COUNTRY_LANG_VARIANTS = {
-    'HK': 'zh_Hant'
+    'HK': 'zh_Hant',
     'TW': 'zh_Hant'
 };
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -2,7 +2,6 @@
 import type { LocaleType } from '@paypal/sdk-constants/src/';
 import { COUNTRY_LANGS, LANG } from '@paypal/sdk-constants/src';
 
-
 export function getComputedLocales(locale : string) : LocaleType {
     let [ lang, country ] = locale.split('_');
     // $FlowFixMe
@@ -11,7 +10,6 @@ export function getComputedLocales(locale : string) : LocaleType {
     if (countryLangs.includes(LANG.ZH_HANT)) {
         lang = LANG.ZH_HANT;
     }
-
 
     // $FlowFixMe
     return { lang, country };
--- a/src/utils.js
+++ b/src/utils.js
@@ -1,16 +1,18 @@
 /* @flow */
-import type { LocaleType } from '@paypal/sdk-constants/src';
+import type { LocaleType } from '@paypal/sdk-constants/src/';
+import { COUNTRY_LANGS } from '@paypal/sdk-constants/src';
 
-const SPECIAL_COUNTRY_LANG_VARIANTS = {
-    'HK': 'zh_Hant',
-    'TW': 'zh_Hant'
-};
 
 export function getComputedLocales(locale : string) : LocaleType {
     let [ lang, country ] = locale.split('_');
-    if (country in SPECIAL_COUNTRY_LANG_VARIANTS) {
-        lang = SPECIAL_COUNTRY_LANG_VARIANTS[country];
+    // $FlowFixMe
+    const countryLangs = COUNTRY_LANGS[country];
+
+    if (countryLangs.includes('zh_Hant')) {
+        lang = 'zh_Hant';
     }
+
+
     // $FlowFixMe
     return { lang, country };
 }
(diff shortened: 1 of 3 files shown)


mimckenna (commented):
This solution will work for other PayPal apps using _paypal-sdk-client_ .  My concern is that it will make no change to [paypal-sdk-constants/src/locale.js](https://github.com/paypal/paypal-sdk-constants/blob/master/src/locale.js).  Several other apps use the `COUNTRY_LANGS` array as is without the help of _paypal-sdk-client_ .  So apps such as checkout-components resolves which language to use for buttons ([link](https://github.com/paypal/paypal-checkout-components/blob/master/src/ui/buttons/props.js#L483-L485)) , it won't find the content.

If this is synchronized with your other PRs noted above, this should work.

We should check the impact of other apps [relying on COUNTRY_LANGS](https://github.com/search?q=org%3Apaypal+COUNTRY_LANGS&type=code), specifically
* [paypal-installments](https://github.com/paypal/paypal-installments/blob/master/server/params.js) - this _should_ work as it has no content of it's own and pulls everything from CDN
* [paypal-smart-payment-buttons](https://github.com/paypal/paypal-smart-payment-buttons/search?q=COUNTRY_LANGS) - smart payment buttons _should_ work since all content is pulled from CDN.  But we'll need to verify that this change doesn't break it.

src/utils.js
```diff
@@ -0,0 +1,15 @@
+/* @flow */
+import type { LocaleType } from '@paypal/sdk-constants/src';
+
+const SPECIAL_COUNTRY_LANG_VARIANTS = {
+    'HK': 'zh_Hant'
```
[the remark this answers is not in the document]
gabrielo91: Added! and yes, this change seems to be a big deal, I just updated the PR, but let me try all together tomorrow, I'll share with you how I test all of this in order to validate it and make sure that I'm covering the needed cases properly.

Regarding to your previous comment, I'm not sure, specially because we are only having into account the special countries [here](https://github.com/paypal/paypal-sdk-client/pull/79/files#diff-978bff6f08d9f62278251257f4d62e5bf7e06dd48cf6db4bca16c37875a28e64R214) (`paypal-sdk-client`) .. however, there is a similar function [here](https://github.paypal.com/Checkout-R/clientsdknodeweb/blob/5eef78655ff88e21c1c90948c3cdac4d3c70f831/server/sdk/params.js#L156) in `clientsdknodeweb` . So, I need to duplicate the `getComputedLocales` function in both projects, but honestly, the scope of this change is a little blurry..  I'm going to do more tests and I'll let u know

src/utils.js
```diff
@@ -0,0 +1,16 @@
...
```
mimckenna, line 5: Are you missing a comma here?

gabrielo91: my bad, solved!

mimckenna (approved):
Looks good now 👍

fedperezp (approved):

src/utils.js
```diff
@@ -0,0 +1,16 @@
+/* @flow */
+import type { LocaleType } from '@paypal/sdk-constants/src';
+
+const SPECIAL_COUNTRY_LANG_VARIANTS = {
+    'HK': 'zh_Hant',
+    'TW': 'zh_Hant'
+};
```
gregjopa, line 7: I worry about maintaining another list here. What do you think about reducing this `COUNTRY_LANGS` from sdk-constant to get all the countries that support `zh_Hant`? 

https://github.com/paypal/paypal-sdk-constants/blob/master/src/locale.js#L460

westeezy: That seems like a more maintainable approach to me as well. Then we can maintain the list in a singular spot which could prove useful later.

gabrielo91: I was doing some tests with that approach, we change the logic a little bit, in this new case, if the list of languages for the country includes `zh_Hant` , then return it as lang ... what do u think?

```
/* @flow */
import type { LocaleType } from '@paypal/sdk-constants/src/';
import { COUNTRY_LANGS } from '@paypal/sdk-constants/src';


export function getComputedLocales(locale : string) : LocaleType {
    let [ lang, country ] = locale.split('_');
    // $FlowFixMe
    const countryLangs = COUNTRY_LANGS[country];

    if (countryLangs.includes('zh_Hant')) {
        lang = 'zh_Hant';
    }


    // $FlowFixMe
    return { lang, country };
}
```

src/utils.js
```diff
@@ -0,0 +1,18 @@
+/* @flow */
+import type { LocaleType } from '@paypal/sdk-constants/src/';
+import { COUNTRY_LANGS } from '@paypal/sdk-constants/src';
```
gregjopa, line 3:
```suggestion
import { COUNTRY_LANGS, LANG } from '@paypal/sdk-constants/src';
```

src/utils.js
```diff
@@ -0,0 +1,18 @@
...
+
+
+export function getComputedLocales(locale : string) : LocaleType {
+    let [ lang, country ] = locale.split('_');
+    // $FlowFixMe
+    const countryLangs = COUNTRY_LANGS[country];
+
+    if (countryLangs.includes('zh_Hant')) {
```
gregjopa, line 11: Nice improvements! Can we also use `LANG.ZH_HANT` from sdk-constants?
```suggestion
    if (countryLangs.includes(LANG.ZH_HANT)) {
```

src/utils.js
```diff
@@ -0,0 +1,18 @@
...
+        lang = 'zh_Hant';
```
gregjopa, line 12:
```suggestion
        lang = LANG.ZH_HANT;
```

src/utils.js
gregjopa: I like this approach! I left a couple comments. 

Can you also add a test for this use case?

src/utils.js
gabrielo91: Nice suggestion, thanks!

src/utils.js
```diff
@@ -0,0 +1,18 @@
+/* @flow */
+import type { LocaleType } from '@paypal/sdk-constants/src/';
+import { COUNTRY_LANGS, LANG } from '@paypal/sdk-constants/src';
+
+
```
gregjopa, line 5: Nit: let's use one blank line to separate code blocks.
```suggestion
```

src/utils.js
```diff
@@ -0,0 +1,18 @@
...
+export function getComputedLocales(locale : string) : LocaleType {
+    let [ lang, country ] = locale.split('_');
+    // $FlowFixMe
+    const countryLangs = COUNTRY_LANGS[country];
+
+    if (countryLangs.includes(LANG.ZH_HANT)) {
+        lang = LANG.ZH_HANT;
+    }
+
+
```
gregjopa, line 15:
```suggestion
```

src/utils.js
```diff
@@ -0,0 +1,18 @@
...
+    // $FlowFixMe
```
gregjopa, line 16: I noticed you have two `$FlowFixMe` comments in here. What flow errors are you getting? It would be great to try and eliminate those and only use them as a last resort.

gabrielo91: I thought it would be more difficult to solve them but I found a way, thanks!!

src/utils.js
```diff
@@ -0,0 +1,17 @@
+/* @flow */
+import type { LocaleType } from '@paypal/sdk-constants/src/';
+import { COUNTRY_LANGS, LANG, COUNTRY } from '@paypal/sdk-constants/src';
+
+export function getComputedLocales(locale : string) : LocaleType {
+    const [ _lang, _country ] = locale.split('_');
```
gregjopa, line 6: I think we can avoid the extra variables by using `let` here. Ex:
```js
let [ lang, country ] = locale.split('_');

lang = LANG[lang.toUpperCase()];
country = COUNTRY[country];
const countryLangs = COUNTRY_LANGS[country];
```

src/utils.js
gabrielo91: Sure, added!

src/utils.js
gabrielo91: my bad, fixed, thanks!

test/client/script.js
```diff
@@ -409,4 +409,20 @@ describe(`script cases`, () => {
             throw new Error(`Expected locale to be ${ expectedLocale }, got ${ receivedLocale }`);
         }
     });
+
+    it('should return computed locale when lang is zh_Hant', () => {
+        const expectedLocale = 'zh_Hant';
```
gregjopa, line 414:
```suggestion
        const expectedLang = 'zh_Hant';
```

test/client/script.js
```diff
@@ -409,4 +409,20 @@ describe(`script cases`, () => {
...
+
+        const url = insertMockSDKScript({
+            query: {
+                'locale': 'zh_HK'
+            }
+        });
+
+        const localeObject = getLocale();
```
gregjopa, line 422:
```suggestion
        const { lang: expectedLang } = getLocale();
```

gregjopa: Please use `expectedLang`/`receivedLang` for naming.

gabrielo91: improved, thanks!

gregjopa (approved):
Looks great! 💯

PR #86, opened 2021-10-21
Add failing test for merchant-id emails

gregjopa:
The `merchant-id` query string parameter accepts an email address or an id. We have a support ticket for adding support for email addresses containing characters like `+`: `'test1+test2@foo.com'`. Currently this email address throws a validation error: https://www.paypal.com/sdk/js?client-id=sb&merchant-id=test1+test2@foo.com

I researched this a bit and it looks like using `url.parse(urlString, true)` drops the `+` causing it to be `'test1 test2@foo.com'`. 

Perhaps we can use a custom escape function with `querystring.parse()` to best solve. https://nodejs.org/api/querystring.html#querystringparsestr-sep-eq-options

### DIFF
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -15,6 +15,9 @@ type SDKMeta = {|
     getSDKLoader : (options? : {| baseURL? : string, nonce? : string |}) => string
 |};
 
+const emailRegex = /^.+@.+$/;
+const emailMaxLengthRegex = /^.{1,64}$/
+
 function validatePaymentsSDKUrl({ pathname, query, hash }) {
 
     if (pathname !== SDK_PATH) {
@@ -37,9 +40,22 @@ function validatePaymentsSDKUrl({ pathname, query, hash }) {
             throw new TypeError(`Unexpected non-string key for sdk url: ${ key }`);
         }
 
-        if (!val.match(/^[a-zA-Z0-9_,-@.]+$/) && !val.match(/^\*$/)) {
+        if (!val.match(/^[a-zA-Z0-9+_,-@.]+$/) && !val.match(/^\*$/)) {
             throw new Error(`Unexpected characters in query key for sdk url: ${ key }=${ val }`);
         }
+
+        if (key === SDK_QUERY_KEYS.MERCHANT_ID) {
+            const merchantValues = val.split(",");
+            Array.from(merchantValues).forEach(aMerchantValue => {
+                if (!emailMaxLengthRegex.test(aMerchantValue)) {
+                    throw new Error(`Email is too long: ${aMerchantValue}`)
+                }
+                if (aMerchantValue.includes("@") && !emailRegex.test(aMerchantValue)) {
+                    throw new Error(`Malformed. merchant email: ${aMerchantValue}`);
+                }
+            });
+        }
+
     }
 
     if (hash) {
--- a/test/server/meta.test.js
+++ b/test/server/meta.test.js
@@ -218,6 +218,56 @@ test('should unpack a valid sdk meta bundle with multiple components', () => {
     }
 });
 
+test('should unpack a valid sdk meta bundle with multiple merchant-id email addresses', () => {
+    const emails = [
+        'test@gmail.com',
+        'foo@bar.com',
+        'test@test.org.uk',
+        'test-test@test.com',
+        'test.test@test.com',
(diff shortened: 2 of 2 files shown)


test/server/meta.test.js
```diff
@@ -218,6 +218,60 @@ test('should unpack a valid sdk meta bundle with multiple components', () => {
...
 });
 
+test('should unpack a valid sdk meta bundle with multiple merchant-id email addresses', () => {
+    const emails = [
+        'test@gmail.com',
+        'foo@bar.com',
+        'test@test.org.uk',
+        'test-test@test.com',
+        'test.test@test.com',
+        'test%2Btest@test.com'
```
gregjopa, line 228: I wonder if we can assume that `'test+test@test.com` will come over encoded at this point in the flow to `'test%2Btest@test.com'`. Testing with clientsdknodeweb should help inform.

server/meta.jsx
```diff
@@ -37,9 +39,19 @@ function validatePaymentsSDKUrl({ pathname, query, hash }) {
             throw new TypeError(`Unexpected non-string key for sdk url: ${ key }`);
         }
 
-        if (!val.match(/^[a-zA-Z0-9_,-@.]+$/) && !val.match(/^\*$/)) {
+        if (!val.match(/^[a-zA-Z0-9+_,-@.]+$/) && !val.match(/^\*$/)) {
```
gregjopa, line 42: Can we only allow `+` for merchantID?

server/meta.jsx
```diff
@@ -15,6 +15,8 @@ type SDKMeta = {|
     getSDKLoader : (options? : {| baseURL? : string, nonce? : string |}) => string
 |};
 
+const emailRegex = /^(([^<>()[\]\.,;:\s@\"]+(\.[^<>()[\]\.,;:\s@\"]+)*)|(\".+\"))@(([^<>()[\]\.,;:\s@\"]+\.)+[^<>()[\]\.,;:\s@\"]{2,})$/i;
```
bluepnume, line 18: Recommend we don't go this path. For emails, it's best just to match `.+@.+` and limit to a length of <50 characters or so.

server/meta.jsx
```diff
@@ -95,7 +107,11 @@ function isLocalUrl(host : string) : boolean {
 }
 
 function validateSDKUrl(sdkUrl : string) {
-    const { protocol, host, hostname, pathname, query, hash } = urlLib.parse(sdkUrl, true);
+    const processedUrl = sdkUrl.replace(/merchant-id=([^\&$]+)/, (_val, match) => {
```
bluepnume, line 110: I'm really confused about what is happening here. It is the merchant's job to correctly encode the url, why are we doing it here?

test/server/meta.test.js
```diff
@@ -218,6 +218,60 @@ test('should unpack a valid sdk meta bundle with multiple components', () => {
...
+    const emails = [
+        'test@gmail.com',
+        'foo@bar.com',
+        'test@test.org.uk',
+        'test-test@test.com',
+        'test.test@test.com',
+        'test%2Btest@test.com',
+    ];
+
+    const sdkUrl = `https://www.paypal.com/sdk/js?client-id=foo&merchant-id=${ emails.join(',') }`;
```
gregjopa, line 231: I would prefer to list the email addresses in their unencoded format and then encode it later with `encodeURIComponent`. Can you give this a try?

```js
const emails = [
        'test@gmail.com',
        'foo@bar.com',
        'test@test.org.uk',
        'test-test@test.com',
        'test.test@test.com',
        'test+test@test.com'
];

const encodedEmails = emails.map(email => encodeURIComponent(email)).join(',');
const sdkUrl = `https://www.paypal.com/sdk/js?client-id=foo&merchant-id=${ encodedEmails }`;
```

server/meta.jsx
```diff
@@ -15,6 +15,9 @@ type SDKMeta = {|
     getSDKLoader : (options? : {| baseURL? : string, nonce? : string |}) => string
 |};
 
+const emailRegex = /^.+@.+$/;
```
westeezy, line 18: Should we use a more robust email regex? Or are we just shooting for a rough match?

westeezy: whoops sorry just saw this was closed 😆. disregard

PR #93, opened 2021-11-18
Eslint rule for compatibility with IE11

almeidacavalcante:
Creating an ESLint rule for not permiting the function `Array.includes()` inside the folder `src/`. Notice that `/server` and `/test` are not affected by this rule.

This change is needed for compatibility with IE11 (to avoid future problems)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,9 @@
 
 module.exports = {
     'extends': './node_modules/grumbler-scripts/config/.eslintrc-browser.js',
-
+    rules: {
+        'dont-use-includes': 'error',
+    },
     globals: {
         __STAGE__: true,
         __VERSION__: true,
--- a/lint-rules/dont-use-includes.js
+++ b/lint-rules/dont-use-includes.js
@@ -1,0 +1,14 @@
+module.exports = {
+    create(context) {
+        return {
+            CallExpression(node) {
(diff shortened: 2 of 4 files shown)


crookedneighbor:
Is there a reason this is going here rather than in the client config in grumblr scripts?

almeidacavalcante:
> Is there a reason this is going here rather than in the client config in grumblr scripts?

Hey, @crookedneighbor, it is my first time on this project. I'm not aware of this configuration you cited. Can you elaborete more? Thanks

crookedneighbor:
I believe the eslint config lives here: https://github.com/krakenjs/grumbler-scripts/blob/master/config/.eslintrc-browser.js

And I would have expected a rule like this to apply to most projects using it, not just the paypal-sdk-client repo

crookedneighbor:
I could be wrong though! There's some ambiguity about what needs to support IE11 and what does not.

westeezy:
Yes, I agree if we are going to change lint rules we should be changing the browser config in grumbler-scripts as @crookedneighbor linked. I have been working in that repo lately to try and modernize it a bit so feel free to reach out with questions related to it. It is essentially a repo of shared configs for the SDK.

almeidacavalcante:
Well, thank you for the informations. I'll take a look at this project. 👍

PR #95, opened 2021-11-19
(DTPPSDK-424) better merchant-id email validation

asquiroga:
Related Draft PR: https://github.com/paypal/paypal-sdk-client/pull/86

Related ticket: https://engineering.paypalcorp.com/jira/browse/DTPPSDK-424

### DIFF
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -16,7 +16,7 @@ type SDKMeta = {|
 |};
 
 const emailRegex = /^.+@.+$/;
-const emailMaxLengthRegex = /^.{1,64}$/
+const emailMaxLengthRegex = /^.{1,256}$/
 
 function validatePaymentsSDKUrl({ pathname, query, hash }) {
 
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -16,7 +16,6 @@ type SDKMeta = {|
 |};
 
 const emailRegex = /^.+@.+$/;
-const emailMaxLengthRegex = /^.{1,256}$/
 
 function validatePaymentsSDKUrl({ pathname, query, hash }) {
 
@@ -47,7 +46,7 @@ function validatePaymentsSDKUrl({ pathname, query, hash }) {
         if (key === SDK_QUERY_KEYS.MERCHANT_ID) {
             const merchantValues = val.split(",");
             merchantValues.forEach(merchantValue => {
-                if (!emailMaxLengthRegex.test(merchantValue)) {
+                if (merchantValue.length > 320) {
                     throw new Error(`Email is too long: ${merchantValue}`)
                 }
                 if (!emailRegex.test(merchantValue)) {
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -15,6 +15,9 @@ type SDKMeta = {|
     getSDKLoader : (options? : {| baseURL? : string, nonce? : string |}) => string
 |};
 
+const emailRegex = /^.+@.+$/;
+const emailMaxLengthRegex = /^.{1,64}$/
+
 function validatePaymentsSDKUrl({ pathname, query, hash }) {
 
     if (pathname !== SDK_PATH) {
@@ -37,9 +40,22 @@ function validatePaymentsSDKUrl({ pathname, query, hash }) {
             throw new TypeError(`Unexpected non-string key for sdk url: ${ key }`);
         }
 
-        if (!val.match(/^[a-zA-Z0-9_,-@.]+$/) && !val.match(/^\*$/)) {
+        if (!val.match(/^[a-zA-Z0-9+_,-@.]+$/) && !val.match(/^\*$/)) {
             throw new Error(`Unexpected characters in query key for sdk url: ${ key }=${ val }`);
         }
+
+        if (key === SDK_QUERY_KEYS.MERCHANT_ID) {
+            const merchantValues = val.split(",");
+            Array.from(merchantValues).forEach(aMerchantValue => {
+                if (!emailMaxLengthRegex.test(aMerchantValue)) {
+                    throw new Error(`Email is too long: ${aMerchantValue}`)
+                }
+                if (aMerchantValue.includes("@") && !emailRegex.test(aMerchantValue)) {
+                    throw new Error(`Malformed. merchant email: ${aMerchantValue}`);
+                }
+            });
+        }
+
     }
 
     if (hash) {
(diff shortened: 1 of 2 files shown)


server/meta.jsx
```diff
@@ -15,6 +15,9 @@ type SDKMeta = {|
     getSDKLoader : (options? : {| baseURL? : string, nonce? : string |}) => string
 |};
 
+const emailRegex = /^.+@.+$/;
```
westeezy, line 18: are we just looking for "roughly like an email" or do we want to use an RFC compliant regex?

server/meta.jsx
```diff
@@ -37,9 +40,22 @@ function validatePaymentsSDKUrl({ pathname, query, hash }) {
...
             throw new Error(`Unexpected characters in query key for sdk url: ${ key }=${ val }`);
         }
+
+        if (key === SDK_QUERY_KEYS.MERCHANT_ID) {
+            const merchantValues = val.split(",");
+            Array.from(merchantValues).forEach(aMerchantValue => {
+                if (!emailMaxLengthRegex.test(aMerchantValue)) {
+                    throw new Error(`Email is too long: ${aMerchantValue}`)
+                }
+                if (aMerchantValue.includes("@") && !emailRegex.test(aMerchantValue)) {
```
westeezy, line 53: if it gets here wouldnt it have already passed the regex which requires `@`?

server/meta.jsx
westeezy: Talked a bit offline. Marking resolved.

server/meta.jsx
gregjopa: Good point @westeezy. We don't need to check for `@` twice. 
Screen Shot 2021-11-29 at 1 36 58 PM
e.

@asquiroga can you update it so it only has one check? My initial thought is to get rid of `aMerchantValue.includes("@")`.

gregjopa, line 53:
```suggestion
                if (!emailRegex.test(aMerchantValue)) {
```

server/meta.jsx
```diff
@@ -15,6 +15,9 @@ type SDKMeta = {|
     getSDKLoader : (options? : {| baseURL? : string, nonce? : string |}) => string
 |};
 
+const emailRegex = /^.+@.+$/;
+const emailMaxLengthRegex = /^.{1,64}$/
```
gregjopa, line 19: I thought more about this and would prefer a more relaxed length restriction. 
```suggestion
const emailMaxLengthRegex = /^.{1,256}$/
```

server/meta.jsx
asquiroga: thank you! commited

server/meta.jsx
```diff
@@ -37,9 +40,22 @@ function validatePaymentsSDKUrl({ pathname, query, hash }) {
...
         }
 
-        if (!val.match(/^[a-zA-Z0-9_,-@.]+$/) && !val.match(/^\*$/)) {
+        if (!val.match(/^[a-zA-Z0-9+_,-@.]+$/) && !val.match(/^\*$/)) {
             throw new Error(`Unexpected characters in query key for sdk url: ${ key }=${ val }`);
         }
+
+        if (key === SDK_QUERY_KEYS.MERCHANT_ID) {
+            const merchantValues = val.split(",");
+            Array.from(merchantValues).forEach(aMerchantValue => {
```
gregjopa, line 49: Do we need the Array.from() call? I think we can remove it since `split()` will always return an array.
```suggestion
            merchantValues.forEach(aMerchantValue => {
```

gregjopa: Also can we rename `aMerchantValue` to just `merchantValue`?

wsbrunson (approved):

server/meta.jsx
```diff
@@ -37,9 +40,22 @@ function validatePaymentsSDKUrl({ pathname, query, hash }) {
...
 
-        if (!val.match(/^[a-zA-Z0-9_,-@.]+$/) && !val.match(/^\*$/)) {
+        if (!val.match(/^[a-zA-Z0-9+_,-@.]+$/) && !val.match(/^\*$/)) {
             throw new Error(`Unexpected characters in query key for sdk url: ${ key }=${ val }`);
         }
+
+        if (key === SDK_QUERY_KEYS.MERCHANT_ID) {
+            const merchantValues = val.split(",");
+            merchantValues.forEach(merchantValue => {
+                if (!emailMaxLengthRegex.test(merchantValue)) {
```
crookedneighbor, line 50: is there a reason to use a regex here? Wouldn't this be simpler (and I assume more performant):

```suggestion
                if (merchantValue.length > 256) {
```

Though I assume you'd pull out the `256` into a `const` at the top like `const maxEmailLength = 256` or something.

Though, also, this suggests that the max email address length is 320: https://www.rfc-editor.org/errata/eid1690#:~:text=It%20should%20say%3A-,In%20addition%20to%20restrictions%20on%20syntax%2C%20there%20is%20a%20length,total%20length%20of%20320%20characters.

test/server/meta.test.js
```diff
@@ -218,6 +218,56 @@ test('should unpack a valid sdk meta bundle with multiple components', () => {
...
+        throw new Error(`Expected script url to be ${ sdkUrl } - got ${ src }`);
+    }
+});
+
+test('should error out from invalid merchant-id email addresses', () => {
+    const emails = [
+        '@',
+        '@io',
+        '@test.com',
+        `${ 'a-very-long-email'.repeat(15) }@a-very-long-domain.com`
```
crookedneighbor, line 250: is there a reason this isn't a separate test? Since long emails throw a different error, I'd expect them to be unit tested separately from these other ones that all test the same path.

test/server/meta.test.js
```diff
@@ -218,6 +218,56 @@ test('should unpack a valid sdk meta bundle with multiple components', () => {
...
+    if (src !== sdkUrl) {
+        throw new Error(`Expected script url to be ${ sdkUrl } - got ${ src }`);
+    }
+});
+
+test('should error out from invalid merchant-id email addresses', () => {
+    const emails = [
+        '@',
+        '@io',
+        '@test.com',
```
crookedneighbor, line 249: Should these cases also be tested:

```suggestion
        '@test.com',
        'name@',
        'no_at_sign',
```

gregjopa:
Great suggestions @crookedneighbor! @asquiroga can you work on these recommended changes?

test/server/meta.test.js
asquiroga: They were all grouped in the same test because all of them are errors related to email-address. But it is a good idea, I will use another test

test/server/meta.test.js
asquiroga: done

gregjopa (approved):

PR #98, opened 2021-12-13
Enable global variable to force Logger URL set to PayPal domain

borodovisin:
### Description
This PR contains a new global variable to force Logger to use the PayPal domain and not use the actual domain when `__TEST__` and `__WEB__` variables are `true`. The source of the problem is this function `getPayPalLoggerUrl` after called `buildPayPalUrl`. If the global variables `__TEST__` and `__WEB__` are `true` the Logger will use the current domain all the time. This is a new requirement after the`Geo` integration.

Please check out this Jira ticket form for more information: [JIRA](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-514)

### Reproduction Steps
This Jira ticket contains an explanation of the problem: [JIRA](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-514)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,7 +20,7 @@ module.exports = {
         __COMPONENTS__: true,
         __FUNDING_ELIGIBILITY__: true,
         __PAYPAL_DOMAIN__: true,
-        __FORCE_PAYPAL_DOMAIN__: true,
+        __FORCE_PAYPAL_DOMAIN_LOGGER__: true,
         __PAYPAL_API_DOMAIN__: true
     }
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
         __COMPONENTS__: true,
         __FUNDING_ELIGIBILITY__: true,
         __PAYPAL_DOMAIN__: true,
+        __FORCE_PAYPAL_DOMAIN__: true,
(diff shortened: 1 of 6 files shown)


src/domains.js
```diff
@@ -37,7 +37,9 @@ export function buildPayPalAPIUrl(path : string = '') : string {
 }
 
 export function getPayPalLoggerUrl() : string {
-    return buildPayPalUrl(URI.LOGGER);
+    return __FORCE_PAYPAL_DOMAIN_LOGGER__ ?
```
borodovisin, line 40: Modified code here to not affect the logic on `buildPayPalUrl`. Not sure if do we need to specialize the code a little more.

.eslintrc.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
         __COMPONENTS__: true,
         __FUNDING_ELIGIBILITY__: true,
         __PAYPAL_DOMAIN__: true,
+        __FORCE_PAYPAL_DOMAIN_LOGGER__: true,
```
borodovisin, line 23: New global variable to force logger into PayPal domain. Not sure how this library is deployed. I guess we can make this change for the China team. They have an odd integration case, only applicable to them.

borodovisin:
This PR do not solve the problem with Logger on different domains

PR #99, opened 2022-01-11
Add data-popups-disabled attribute

ravishekhar:
Add support for data-popups-disabled attribute when loading JS SDK.


src/script.js
```diff
@@ -200,6 +200,10 @@ export function getPartnerAttributionID() : ?string {
     return getSDKAttribute(SDK_SETTINGS.PARTNER_ATTRIBUTION_ID);
 }
 
+export function getMerchantRequestedPopupDisabled() : boolean {
```
gregjopa, line 203: I wonder if we need this new function or if checkout-components can just use `getSDKAttribute()` directly instead.

ravishekhar: This is added mainly for consistency to follow the current pattern of accessing all attributes via a helper function.

gregjopa (approved):

server/meta.jsx
```diff
@@ -140,6 +140,7 @@ const ALLOWED_ATTRS = [
     SDK_SETTINGS.CLIENT_TOKEN,
     SDK_SETTINGS.MERCHANT_ID,
     SDK_SETTINGS.PARTNER_ATTRIBUTION_ID,
+    SDK_SETTINGS.DATA_POPUPS_DISABLED,
```
gregjopa, line 143: I removed the "DATA_" part of this name in sdk-constants since it was the only setting using it.
```suggestion
    SDK_SETTINGS.POPUPS_DISABLED,
```

src/meta.js
```diff
@@ -9,6 +9,7 @@ const ALLOWED_ATTRS = [
     SDK_SETTINGS.AMOUNT,
     SDK_SETTINGS.MERCHANT_ID,
     SDK_SETTINGS.PARTNER_ATTRIBUTION_ID,
+    SDK_SETTINGS.DATA_POPUPS_DISABLED,
```
gregjopa, line 12:
```suggestion
    SDK_SETTINGS.POPUPS_DISABLED,
```

src/script.js
gregjopa, line 203: I'm thinking we make `Popups` plural to align with the `POPUPS_DISABLED` constant

```suggestion
export function getMerchantRequestedPopupsDisabled() : boolean {
```

src/script.js
```diff
@@ -200,6 +200,10 @@ export function getPartnerAttributionID() : ?string {
...
+    return getSDKAttribute(SDK_SETTINGS.DATA_POPUPS_DISABLED) === 'true';
```
gregjopa, line 204:
```suggestion
    return getSDKAttribute(SDK_SETTINGS.POPUPS_DISABLED) === 'true';
```

ravishekhar (commented):
Use plural form of popups in the name and remove DATA_ from the constants

gregjopa (approved):

test/client/script.js
```diff
@@ -3,7 +3,7 @@
 import { base64encode } from 'belter/src';
 
 import { getClientID, getIntent, getCurrency, getVault, getCommit, getClientToken, getPartnerAttributionID,
-    getMerchantID, getClientAccessToken, getSDKIntegrationSource, insertMockSDKScript, getPageType, getLocale } from '../../src';
+    getMerchantID, getClientAccessToken, getSDKIntegrationSource, insertMockSDKScript, getPageType, getLocale, getMerchantRequestedPopupDisabled } from '../../src';
```
gregjopa:
```suggestion
    getMerchantID, getClientAccessToken, getSDKIntegrationSource, insertMockSDKScript, getPageType, getLocale, getMerchantRequestedPopupsDisabled } from '../../src';
```

ravishekhar: Applied in a separate commit.

PR #101, opened 2022-01-14
add support for browser locales on getBuyerCountry

agnjunio:
This add SPB the ability to infer the country from the user's browser language settings. By improving country detection, we  perform better eligibility checks using the client browser settings.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -165,7 +165,30 @@ export function getDisableCard() : $ReadOnlyArray<?$Values<typeof CARD>> {
 }
 
 export function getBuyerCountry() : ?$Values<typeof COUNTRY> {
-    return getSDKQueryParam(SDK_QUERY_KEYS.BUYER_COUNTRY);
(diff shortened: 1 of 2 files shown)


gregjopa (commented):
The `SDK_QUERY_KEYS.BUYER_COUNTRY` query param is only designed for Sandbox so merchants can test out different flows based on buyer country. Is this locale buyer country change for Sandbox only or both Sandbox and production?

In production, the merchant cannot override buyer country and PayPal's router is responsible for setting Buyer Country in the request header and server-side apps can read this header.

gregjopa:
Hi @agnjunio, thanks for trying to help improve buyer country detection. This `getBuyerCountry()` function in sdk-client is designed for sandbox testing only. It's used to let merchants test out different buyer experiences in sandbox. It production we have different rules for how we get buyerCountry and we do not allow overriding it from the client-side. 

Closing this out since sdk-client is not the right spot for this buyer country/locale change.

PR #103, opened 2022-01-24
Improve domain validation logic

gregjopa:
This PR improves the `validateSDKUrl()` server-side logic with the following changes:
1. Ensures the protocol really starts with http or https. The Node url.parse() function seems to ignore certain prefixes. In the screenshot below, note how the original url string starts with `/uFEFF` but the parsed protocol  shows `https:`
Screen Shot 2022-01-24 at 4 31 27 PM

2. Prevent backslashes and other potentially harmful characters from being included in the subdomain of the url.

### DIFF
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -80,10 +80,6 @@ function isLegacySDKUrl(hostname : string, pathname : string) : boolean {
 }
 
 function isSDKUrl(hostname : string) : boolean {
-    // if (hostname.match(/https:\/\/[a-z0-9\.]+.paypal.(com|cn)/)) {
-    //     return true;
-    // }
-
     if (hostname.endsWith(HOST.PAYPAL) || hostname.endsWith(HOST.PAYPAL_CHINA)) {
         return true;
     }
@@ -111,6 +107,12 @@ function validateSDKUrl(sdkUrl : string) {
 
     if (!sdkUrl.startsWith(PROTOCOL.HTTP) && !sdkUrl.startsWith(PROTOCOL.HTTPS)) {
         throw new Error(`Expected protocol for sdk url to be ${ PROTOCOL.HTTP } or ${ PROTOCOL.HTTPS } for host: ${ hostname } - got ${ protocol || 'undefined' }`);
+    }
+
+    const hostnameMatchResults = hostname.match(/[a-z0-9\.\-]+/);
+
+    if (!hostnameMatchResults || hostnameMatchResults[0] !== hostname) {
+        throw new Error(`Expected a valid host: ${ hostname }`);
     }
 
     if (isLegacySDKUrl(hostname, pathname)) {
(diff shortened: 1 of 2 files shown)


test/server/meta.test.js
```diff
@@ -870,3 +870,39 @@ test('should construct a valid script url hosted on www.paypal.cn', () => {
         throw new Error(`Expected script url to be ${ sdkUrl } - got ${ src }`);
     }
 });
+
+test('should error when the script url does not start with "https://"', () => {
```
wsbrunson, line 874: do we already have a test for `http://`?

server/meta.jsx
```diff
@@ -105,10 +105,16 @@ function validateSDKUrl(sdkUrl : string) {
...
 
-    if (protocol !== PROTOCOL.HTTP && protocol !== PROTOCOL.HTTPS) {
+    if (!sdkUrl.startsWith(PROTOCOL.HTTP) && !sdkUrl.startsWith(PROTOCOL.HTTPS)) {
         throw new Error(`Expected protocol for sdk url to be ${ PROTOCOL.HTTP } or ${ PROTOCOL.HTTPS } for host: ${ hostname } - got ${ protocol || 'undefined' }`);
     }
 
+    const hostnameMatchResults = hostname.match(/[a-z0-9\.\-]+/);
+
+    if (!hostnameMatchResults || hostnameMatchResults[0] !== hostname) {
+        throw new Error(`Expected a valid host: ${ hostname }`);
```
wsbrunson, line 115: would it be worthwhile to explain in the error message what is a valid hostname? (only letters, numbers, . and -)

wsbrunson (approved):

test/server/meta.test.js
gregjopa: Great point! I updated the test to check for both `http://` and `:https://`. We have code that technically supports checkout.js on `http://` which is why we support both protocols.

server/meta.jsx
gregjopa: Updated! Great suggestion.

nbierdeman:
I was curious if `hostname` is normalized to only include lower case letters, which would then work well with `hostname.match(/[a-z0-9\.\-]+/)`, and came across this warning against the use of `urlLib.parse()`. Just thought it was worth mentioning (if you see something, say something):

Screen Shot 2022-01-25 at 9 55 09 AM

gregjopa:
Great comments @nbierdeman!! Really appreciate the review 💯   

> I was curious if hostname is normalized to only include lower case letters, which would then work well with hostname.match(/[a-z0-9\.\-]+/)

Yep I'm pretty sure the browser automatically converts any UPPERCASE characters in the domain to lowercase before doing the request. I found this answer on SO:
> Domain names are case insensitive according to RFC 4343. The rest of URL is sent to the server via the GET method. This may be case sensitive or not.

https://stackoverflow.com/a/17113291

> came across this warning against the use of urlLib.parse(). Just thought it was worth mentioning (if you see something, say something):

Great point here. I noticed this awhile ago as well. Ideally we should be using `new URL()` but there are some complexities with switching over since query string parsing logic is different. Definitely doable but is a medium sized effort.

PR #105, opened 2022-01-25
chore: update github actions

gregjopa:


.github/workflows/main.yml
```diff
@@ -33,4 +30,4 @@ jobs:
         run: npm run build
 
       - name: ⬆️ Upload coverage report
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v2
```
gregjopa, line 33: fyi v1 is being deprecated: https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1. We should update this in every repo.

PR #107, opened 2022-02-01
fix injection xss

lordandares:


bluepnume:
Yeah can you share an example of the attack? It looks like you're searching for `{{` and `}}` which makes me think you're trying to prevent the value being used with some template engine?

lordandares:
HI @gregjopa @bluepnume i've tested the new version of the library and the but is still happening, to reproduce this bug:
Add sdkMeta Param To URL: like 
https://www.msmaster.qa.paypal.com/webapps/xoonboarding/?token=EC-62963695CH906051B&sdkMeta=eyJ1cmwiOiJodHRwczovL3d3dy5wYXlwYWwuY29tL3Nkay9qcz9jbGllbnQtaWQ9QVR4WDE2R011V3Q1eG80SGZRdmc5dHFWVEZ3bV9PRW9iaVdaTWp2WXJac05Ta3dNVHBUQzNhYTdWOXZocXpyT3drbzBQV1lEdEx4RlgxbWomY3VycmVuY3k9VVNEIiwiYXR0cnMiOnsiZGF0YS1zZGstaW50ZWdyYXRpb24tc291cmNlIjoiaW5qZWN0aW9uX2hlcmV7e2NvbnN0cnVjdG9yLmNvbnN0cnVjdG9yKCdhbGVydChkb2N1bWVudC5kb21haW4pJykoKX19In19
Replace token if does not load

the sdkMeta is a base64-encoded witn and XSS with an alert

image

gregjopa:
We discussed this internally and decided that the internal Angular apps are the best place to fix it. We should be able to use this Angular bootstrap function to prevent the attack: https://docs.angularjs.org/api/ng/function/angular.bootstrap

PR #108, opened 2022-02-04
Log merchantId from global variable if isn't available from the request parameters

borodovisin:
### Description
The PR adds a new logic to track the `sellerId` on the tracking builder. This means if we don't receive a merchant identifier from the request params will try to get it from the global variables based on the `client-id` sent in the request.

To enable the merchant identifier in the tracking logs after the PayPal buttons initialization. Now only partners integrations sent the `merchantId` in the request params.
More info [here](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-641).

### Dependent Changes
This PR depends on the changes made in repo `clientsdknodeweb`, please check this [PR](https://github.paypal.com/Checkout-R/clientsdknodeweb/pull/323)


wsbrunson (approved):

src/global.js
```diff
@@ -101,3 +101,7 @@ export function getFundingEligibility() : FundingEligibilityType {
 export function getPlatform() : $Values<typeof PLATFORM> {
     return isDevice() ? PLATFORM.MOBILE : PLATFORM.DESKTOP;
 }
+
+export function getMerchantIdFromClientId() : $ReadOnlyArray<string> {
+    return [ __MERCHANT_ID_FROM_CLIENT_ID__ ];
```
bluepnume, line 106: Recommend against this :) this will mean SDK script caches can no longer be shared between different merchants.

borodovisin:
This PR will affect the current behavior of the cache strategy

PR #111, opened 2022-02-10
feat(codecov): enhance the codecov report

borodovisin:
### Description
The PR enhance the codecov report testing coverage creating two different flags, one for the client-side and the other for the server-side

To split the test coverage report target in two paths, the client-side and the server-side

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,26 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 50%
+      client: 
+        target: 50%
+        threshold: 1%
+        flags:
+          - client
+      server:
+        target: 100%
+        threshold: 1%
+        flags:
+          - server
+
+comment:
+  layout: "diff, flags, files"
+
+flags: 
+  client:
+    paths:
+      - src
+  server:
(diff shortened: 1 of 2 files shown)


borodovisin:
I need some definitions here about what are the client test and server test. Also, do we are planning to remove the existing jest and karma flags?
I need to know conceptually how to create the flags to enhance as well the Github actions to publish the flags report and make the check per flags available in the PRs.

wsbrunson (approved):

.github/codecov.yml
```diff
@@ -0,0 +1,26 @@
...
+      server:
+        target: 100%
+        threshold: 1%
+        flags:
+          - server
+
+comment:
+  layout: "diff, flags, files"
+
+flags:
```
gregjopa, line 20: We should be able to remove this flags here. And instead let's rename the existing flags to "client" and "server" here: https://github.com/paypal/paypal-sdk-client/blob/main/.github/workflows/main.yml#L32-L42

borodovisin:
Will start a new PR and reject this one and reuse the existing flags.

PR #113, opened 2022-02-15
Increase jest test to 100% of coverage

borodovisin:
### Description
The PR covert all the missing  Jest tests in the `meta.jsx` file.

To have a full coverage on the server (Jest) unit test

### DIFF
--- a/test/server/meta.test.js
+++ b/test/server/meta.test.js
@@ -425,7 +425,7 @@ test('should error out with an invalid protocol in localhost', () => {
 
 test('should error out with an invalid host', () => {
 
-    const sdkUrl = 'https://www.paypal.example.com/sdk/js?client-id=foo';
+    const sdkUrl = 'https://?client-id=foo';
     let error;
 
     try {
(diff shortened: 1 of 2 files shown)


wsbrunson (approved):

test/server/meta.test.js
```diff
@@ -425,7 +425,7 @@ test('should error out with an invalid protocol in localhost', () => {
 
 test('should error out with an invalid host', () => {
 
-    const sdkUrl = 'https://www.paypal.example.com/sdk/js?client-id=foo';
+    const sdkUrl = 'https:/www.paypal.example.com';
```
gregjopa, line 428: This seems like it would fail the protocol check `https://` instead of the hostname part

test/server/meta.test.js
```diff
@@ -443,7 +443,7 @@ test('should error out with an invalid host', () => {
 
 test('should error out with no path', () => {
 
-    const sdkUrl = 'https://www.paypal.com?client-id=foo';
+    const sdkUrl = 'https://?client-id=foo';
```
gregjopa, line 446:
```suggestion
    const sdkUrl = 'https://www.paypal.com?client-id=foo';
```

test/server/meta.test.js
gregjopa, line 428:
```suggestion
    const sdkUrl = 'https://?client-id=foo';
```

gregjopa (approved):

PR #114, opened 2022-02-16
Test the unpackSDKMeta with real data

borodovisin:
### Description
Create a unit test case with real data information from the logs to make sure the `unpackSDKMeta` is working as expected.

To test the `unpackSDKMeta` with real information from merchants' use cases.

### DIFF
--- a/test/server/meta.integration.test.js
+++ b/test/server/meta.integration.test.js
@@ -1,0 +1,117 @@
+/* @flow */
+
+import cheerio from 'cheerio';
+
+import { unpackSDKMeta } from '../../server';
+
+/**
+ * List with real URL query parameters.
+ * Each value is a tuple with the URL parameters and attributes
+ * The values for all the keys should be a string.
+ */
+const sdkMetaList = [
+    [ {
+        'url':         'https://www.paypal.com/sdk/js?',
+        'client-id':   'AQbghYd-7mRPyimEriYScIgTnYUsLnr5wVnPnmfPaSzwKrUe3qNzfEc5hXr9Ucf_JG_HFAZpJMJYXMuk',
+        'merchant-id': 'K2US3FE3PQUEQ',
+        'intent':      'capture',
+        'locale':      'it_IT',
+        'components':  'buttons,funding-eligibility,messages',
+        'commit':      'false',
+        'currency':    'EUR'
+        
+    }, {
+        'attrs': {
+            'data-partner-attribution-id': 'WooThemes_EC',
+            'data-uid':                    'uid_gwakesmdhfvakwylnuczrbplzdthxe'
+        }
+    } ],
+    [ {
+        'url':         'https://www.paypal.com/sdk/js?',
+        'client-id':   'AfjD8CKlCiagY3OXl46F4HJKHIP7k2wRU3UDfMSuo52DBiXuc0srow-uyWGeHT0EXd67zADzzkm-5eoy',
+        'components':  'buttons',
+        'currency':    'GBP',
+        'intent':      'authorize',
+        'locale':      'en_GB',
+        'merchant-id': 'ASCDJ8AE36AYQ'
+    }, {
+        'attrs': {
+            'data-amount': '7.99',
+            'data-uid':    'uid_issqwuttcumomcdjzbomwiprpnsgjx'
+        }
+    } ],
+    [ {
+        'url':         'https://www.paypal.com/sdk/js?',
+        'client-id':   'AebICI1y5FXWLTk3MHJ-I8g9_lUMlpxf6AP95fnbdtx3WQQFKpMFBP0KsEzeYQAE4GeQ18DlVxvw1qQ9',
+        'commit':      'false',
+        'currency':    'USD',
+        'components':  'buttons,funding-eligibility',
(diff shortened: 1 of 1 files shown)


test/server/meta.test.js
```diff
@@ -1021,3 +1021,60 @@ test('should error when invalid characters are found in the subdomain - we allow
...
+        'url':             'https://www.paypal.com/sdk/js?client-id=AfjD8CKlCiagY3OXl46F4HJKHIP7k2wRU3UDfMSuo52DBiXuc0srow-uyWGeHT0EXd67zADzzkm-5eoy&components=buttons&currency=EUR&intent=authorize&locale=de_DE&merchant-id=3PA5M5QDS5BES',
+        'attrs': {
+            'data-amount': '29.90',
+            'data-uid':    'uid_ssqwuttcumomcdjzbomwiakaomcjzs'
+        }
+    }
+];
+
+// $FlowIgnore[prop-missing] missing each property for test
+test.each(dataSet)('should unpack the sdk meta with same URL', (source) => {
```
gregjopa, line 1071:
```suggestion
test.each(dataSet)('should unpack `sdkMeta` and load the script', (source) => {
```

test/server/meta.test.js
```diff
@@ -1021,3 +1021,60 @@ test('should error when invalid characters are found in the subdomain - we allow
         throw new Error(`Expected error to be thrown`);
     }
 });
+
+const dataSet = [
```
gregjopa, line 1025: What do you think about moving this `dataSet` and test to a separate file? Perhaps we create server/meta.integration.test.js

borodovisin: Sure that sounds good...

test/server/meta.integration.test.js
```diff
@@ -0,0 +1,64 @@
+/* @flow */
+
+import cheerio from 'cheerio';
+
+import { unpackSDKMeta } from '../../server';
+
+const sdkMetaList = [
```
westeezy, line 7: We can potentially simplify the building of theses so its easier to parse when reviewing tests. Right now we would have to really pause and read through each URL which can be difficult and potentially cause duplicate test cases or missed test cases in the long term. How do you feel about something like:
```
const params = [
  [
    {
      clientId: 'AQbghYd-7mRPyimEriYScIgTnYUsLnr5wVnPnmfPaSzwKrUe3qNzfEc5hXr9Ucf_JG_HFAZpJMJYXMuk',
      merchantId: 'K2US3FE3PQUEQ'
      intent: 'capture',
      lcaol: 'it_IT',
      components: ['buttons', 'funding-eligibility', 'messages']
      commit: false,
      currency: 'EUR'
    },
    {
      'data-partner-attribution-id': 'WooThemes_EC',
      'data-uid': 'uid_gwakesmdhfvakwylnuczrbplzdthxe',
    }
  ],
]

const buildTestUrl = ([queryParams, attrs]) => {
   // Build and return the url
  return `todo....`;
}

// $FlowIgnore[prop-missing] missing each property for test
test.each(params)('should unpack `sdkMeta` and load the script', ([queryParams, attrs]) => {
    const source = buildTestUrl([queryParams, attrs]);
    
    const { getSDKLoader } = unpackSDKMeta(Buffer.from(JSON.stringify(source))
        .toString('base64'));

    const $ = cheerio.load(getSDKLoader());
    const src = $('script').attr('src');

    if (source.url !== src) {
        throw new Error('The url should be the same');
    }
});
```

westeezy: doesn't have to be tuples but i jumped there. feel free to use objects of course.

westeezy: One could also make use of the test tables in jest but i kinda find those less readable than above to be honest

borodovisin: That sounds good @westeezy, you're right it's a little hard to understand the object.

wsbrunson (approved):

test/server/meta.integration.test.js
```diff
@@ -0,0 +1,130 @@
...
+];
+
+/**
+ * Generate the URL from the source object
+ *
+ * @param {object} sourceData to generate the URL
+ * @returns {string} the generated URL including all the query parameters
+ */
+function buildUrl(sourceData) : string {
+    const { url, ...queryParam } = sourceData;
```
gregjopa, line 107:
```suggestion
    const { url, ...queryParams } = sourceData;
```

test/server/meta.integration.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Generate the URL from the source object
+ *
+ * @param {object} sourceData to generate the URL
+ * @returns {string} the generated URL including all the query parameters
+ */
+function buildUrl(sourceData) : string {
+    const { url, ...queryParams } = sourceData;
+    const stringQueryParams = Object
```
gregjopa, line 108: Can you leverage `URLSearchParams` to simplify converting an object to a querystring?
Screen Shot 2022-02-17 at 2 27 55 PM

borodovisin: Sure @gregjopa, I'll make the changes

gregjopa (approved):
💯

PR #116, opened 2022-02-23
Sanitize script attributes for angular expressions

ravishekhar:
Sanitize the attributes passed to script tag to remove any angular expressions.


ravishekhar:
Closing this PR as this is not the best place to add this logic.

PR #120, opened 2022-03-04
Increase unit test coverage on domains file

borodovisin:
### Description
The PR increases the unit test coverage in the domain.js file for the client(karma).

### DIFF
--- a/test/client/domains.js
+++ b/test/client/domains.js
@@ -1,10 +1,17 @@
 /* @flow */
+import { ENV } from '@paypal/sdk-constants/src';
 
 import { getPayPalDomainRegex } from '../../src';
+import {
+    getPayPalLoggerDomain,
+    getAuthAPIUrl,
+    getOrderAPIUrl
+} from '../../src/domains';
 
 beforeEach(() => {
     window.__ENV__ = 'test';
 });
+
 describe(`domains test`, () => {
     it('should successfully match valid domain', () => {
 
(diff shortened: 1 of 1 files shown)


westeezy (approved):

test/client/domains.js
```diff
@@ -38,4 +44,47 @@ describe(`domains test`, () => {
...
+        window.__ENV__ = ENV.LOCAL;
+        window.__STAGE_HOST__ = undefined;
+        try {
+            getPayPalLoggerDomain();
+        } catch (err) {
+            if (err.message !== 'No stage host found') {
+                throw new Error(`should thrown exception with message "No stage host found", but got ${ err.message }`);
+            }
+        }
+        window.__ENV__ = previousEnv;
```
gregjopa, line 69: This shouldn't be needed since we clean up `window.__ENV__` in the beforeEach().
```suggestion
```

test/client/domains.js
```diff
@@ -38,4 +44,47 @@ describe(`domains test`, () => {
...
+        window.__STAGE_HOST__ = previousStageHost;
```
gregjopa, line 70: Can you reset this `window.__STAGE_HOST__` in the beforeEach() as well?
```suggestion
```

borodovisin: Not sure why we should move the cleanup to the beforeEach cycle, we modify the `__STAGE_HOST__ ` only for that specific test case, that is the main reason to keep the clean process at the end of that unit test, and not for all the test suite cycle. In any case, if I need to move it to the `beforeEach` let me know...

PR #122, opened 2022-03-08
Increase unit test coverage in the api file 🧪 

borodovisin:
### Description
The PR increases the unit test coverage in the script file for the client(karma). Also, split the scripts test file in two to avoid fat files in the repo.

### Caveat
Take a look at this code: https://github.com/paypal/paypal-sdk-client/blob/52b7f85e19e3c80c65907aecd17f715496e3b0ac/src/api.js#L105 there we check if the `merchantID` is valid, but we get the `merchantID` from the function `getMerchantID`. That specific function returns an empty array if a merchant wasn't passed in the query parameters, meaning the check `![]` always be false in JS, meaning at the end we don't care if the `merchantID` is an empty array or not and we aren't able to throw the exception for the case user define a payee but a merchant id wasn't register corerctly.

### DIFF
--- a/test/client/api.js
+++ b/test/client/api.js
@@ -1,0 +1,164 @@
+/* @flow */
+
+import { $mockEndpoint } from 'sync-browser-mocks/dist/sync-browser-mocks';
+
+import { createAccessToken, createOrder } from '../../src/api';
+
+describe('api cases', () => {
+    const expectedToken = 'A21AAKNZBaqilFBC4dVVz-tr-ySIT78NREeBidy3lkGdr-EA8wbhGrByPayhgnJRPE5xg4QW46moDbCFjZ13i1GH-Ax4SjtjA';
+    const defaultAuthResponse = {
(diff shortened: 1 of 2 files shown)


test/client/api.js
```diff
@@ -0,0 +1,164 @@
...
+            // $FlowIgnore[incompatible-call]
+            await createOrder('testClient', order);
+        } catch (err) {
+            if (err.message !== expectedErrorMessage) {
+                throw new Error(`should throw an error with message '${ expectedErrorMessage }', but got: '${ err.message }'`);
+            }
+        }
+    });
+
+    it('createOrder should throw an error when order currency not match with query parameters currency', async () => {
```
gregjopa, line 120:
```suggestion
    it('createOrder should throw an error when order currency does not match with query parameters currency', async () => {
```

gregjopa (approved):
Great job @borodovisin! I left a few minor comments.

PR #123, opened 2022-03-09
Increase unit test coverage in the graphql file 🧪 

borodovisin:
### Description
The PR increases the unit test coverage in the graphql file for the client(karma).

### DIFF
--- a/src/graphql.js
+++ b/src/graphql.js
@@ -2,7 +2,6 @@
 
 import { ZalgoPromise } from '@krakenjs/zalgo-promise/src';
 import { request, stringifyError } from '@krakenjs/belter/src';
-import { CURRENCY } from '@paypal/sdk-constants/src';
 
 import { getLogger } from './logger';
 import {
@@ -19,23 +18,9 @@ import {
 } from './script';
 import { buildPayPalUrl } from './domains';
 
-
-type FundingEligibilityParams = {|
-    clientID : string,
-    merchantID : ?$ReadOnlyArray<string>,
-    buyerCountry : ?string,
-    currency : $Values<typeof CURRENCY>,
-    commit : boolean,
-    vault : boolean,
-    intent : string,
-    enableFunding : $ReadOnlyArray<?string>,
-    disableFunding : $ReadOnlyArray<?string>,
-    disableCard : $ReadOnlyArray<?string>
-|};
-
 const GRAPHQL_URI = '/graphql';
 
-function buildFundingEligibilityVariables() : FundingEligibilityParams {
+function buildFundingEligibilityVariables() : { [string] : mixed } {
     const clientID = getClientID();
     const merchantID = getMerchantID();
     const buyerCountry = getBuyerCountry();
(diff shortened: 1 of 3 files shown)


wsbrunson (approved):

wsbrunson (approved):

test/client/graphql.js
```diff
@@ -0,0 +1,90 @@
+/* @flow */
+
+import { $mockEndpoint  } from 'sync-browser-mocks/dist/sync-browser-mocks';
```
gregjopa, line 3:
```suggestion
import { $mockEndpoint } from 'sync-browser-mocks/dist/sync-browser-mocks';
```

src/graphql.js
```diff
@@ -61,6 +61,7 @@ function buildFundingEligibilityVariables() : FundingEligibilityParams {
     };
 }
 
+// istanbul ignore next
```
gregjopa, line 64: Can we delete this function in favor of defaulting to an empty object like so?

```js
export function callGraphQL<T, V>({ query, variables = {}, headers = {} } : {| query : string, variables : V, headers? : { [string] : string } |}) : ZalgoPromise<T> {
```

src/graphql.js
```diff
@@ -61,12 +61,7 @@ function buildFundingEligibilityVariables() : FundingEligibilityParams {
     };
 }
 
-function getDefaultVariables<V>() : V {
-    // $FlowFixMe[incompatible-return]
-    return {};
-}
-
-export function callGraphQL<T, V>({ query, variables = getDefaultVariables(), headers = {} } : {| query : string, variables : V, headers? : { [string] : string } |}) : ZalgoPromise<T> {
+export function callGraphQL<T, V = {}>({ query, variables, headers = {} } : {| query : string, variables : V, headers? : { [string] : string } |}) : ZalgoPromise<T> {
```
gregjopa, line 64: Let's maintain the existing behavior where `variables` defaults to an empty object.
```suggestion
export function callGraphQL<T, V = {}>({ query, variables = {}, headers = {} } : {| query : string, variables : V, headers? : { [string] : string } |}) : ZalgoPromise<T> {
```

borodovisin: Hey @gregjopa, I wonder if the `V` type is really a generic type, it seems to me that is a specific type, GraphQl defines how to call the endpoint not the logic, meaning the variables should have a format as well the query argument.
Maybe I'm wrong but I don't see variables as a generic type, we should compliance the GraphQl format.

src/graphql.js
```diff
@@ -61,12 +61,7 @@ function buildFundingEligibilityVariables() : FundingEligibilityParams {
     };
 }
 
-function getDefaultVariables<V>() : V {
-    // $FlowFixMe[incompatible-return]
-    return {};
-}
-
-export function callGraphQL<T, V>({ query, variables = getDefaultVariables(), headers = {} } : {| query : string, variables : V, headers? : { [string] : string } |}) : ZalgoPromise<T> {
+export function callGraphQL<T, V>({ query, variables = {}, headers = {} } : {| query : string, variables : V | { [string] : string }, headers? : { [string] : string } |}) : ZalgoPromise<T> {
```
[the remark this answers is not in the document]
borodovisin: Yeah, that makes sense to me, I going to make the change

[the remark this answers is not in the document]
borodovisin: The type `FundingEligibilityParams` was removed, incompatibility with that type and `mixed` in the value for variables. The removed type is internal we don't export it or the function `buildFundingEligibilityVariables `, I guess is save for us, since anyone else knows about that type.

src/graphql.js
```diff
@@ -19,23 +18,9 @@ import {
 } from './script';
 import { buildPayPalUrl } from './domains';
 
-
-type FundingEligibilityParams = {|
```
gregjopa, line 23: @borodovisin this is an interesting strategy for fixing the type errors 😺 Deleting types and using mixed technically works...

westeezy: Do we expose this type anywhere else? I think we might if so we can just use that. Otherwise I think I'm okay with mixed for the short term. I feel like this should be a type we import in the future.

westeezy (approved):

PR #134, opened 2022-05-06
chore: add failing test for v4

borodovisin:
### Description
Enhance the sdkMeta url


server/meta.jsx
```diff
@@ -192,6 +192,17 @@ function getSDKScriptAttributes(sdkUrl : ?string, allAttrs : ?{ [string] : strin
     return attrs;
 }
 
+function clearQueryFromLegacySDKUrl(sdkUrl : string ) : string {
```
gregjopa, line 195: Could we use the name `sanitize` or `transform`?

```suggestion
function sanitizeSDKUrl(sdkUrl : string ) : string {
if ()


```

server/meta.jsx
```diff
@@ -192,6 +192,17 @@ function getSDKScriptAttributes(sdkUrl : ?string, allAttrs : ?{ [string] : strin
...
+    const { protocol, hostname, port, pathname } = urlLib.parse(sdkUrl);
+
+    return isLegacySDKUrl(String(hostname), String(pathname)) ? urlLib.format({
```
gregjopa, line 198: I don't think we need the `String()` conversion. We don't do that with the other usage of `urlLib.parse()`
```suggestion
    return isLegacySDKUrl(hostname, pathname) ? urlLib.format({
```

server/meta.jsx
```diff
@@ -192,6 +192,18 @@ function getSDKScriptAttributes(sdkUrl : ?string, allAttrs : ?{ [string] : strin
     return attrs;
 }
 
+function sanitizeSDKUrl(sdkUrl : string ) : string {
```
gregjopa, line 195: Minor change request so v5 url does not get manipulated:

```js
function sanitizeSDKUrl(sdkUrl : string ) : string {
    // eslint-disable-next-line compat/compat
    const url = new URL(sdkUrl);

    // remove query string params for checkout.js
    if (isLegacySDKUrl(url.hostname, url.pathname)) {
        url.search = '';
        url.hash = '';

        return url.toString();
    }

    return sdkUrl;
}
```

gregjopa (approved):

wsbrunson (approved):

westeezy (approved):

PR #139, opened 2022-07-28
use same error text for getSDKScript

wsbrunson:
this will help us with functional tests we are writing for clientsdknodeweb


gregjopa (approved):

src/script.js
```diff
@@ -40,6 +40,13 @@ import { getComputedLocales } from "./utils";
 
 type GetSDKScript = () => HTMLScriptElement;
 
+const buildScriptNotFoundError = (host, path, error = "") =>
```
gregjopa, line 43: One thing I noticed while testing `stringifyError("")` is it will return the following for an empty string:
```js
stringifyError("") 
// returns the following
// "<unknown error: [object String]>" 
```

Based on this, should we conditionally include the error results with something like:
```js
 `PayPal Payments SDK script not found on page! Expected to find <script src="https://${host}${path}">\n\n${error && stringifyError(error)}`
```

dtjones404 (approved):

PR #151, opened 2023-05-09
Update domains to add Venmo.com to list of valid domains

ravishekhar:
* Add Venmo.com to the list of valid domains

### DIFF
--- a/src/domains.js
+++ b/src/domains.js
@@ -62,7 +62,7 @@ export function getPayPalDomainRegex(): RegExp {
     return /.*loca.*/;
   }
   // eslint-disable-next-line security/detect-unsafe-regex
-  return /\.paypal\.(com|cn)(:\d+)?$/;
+  return /((\.paypal\.(com|cn))|(\.venmo\.com))(:\d+)?$/;
 }
 
 export function isPayPalDomain(): boolean {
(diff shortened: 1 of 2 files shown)


src/domains.js
```diff
@@ -62,7 +62,7 @@ export function getPayPalDomainRegex(): RegExp {
     return /.*loca.*/;
   }
   // eslint-disable-next-line security/detect-unsafe-regex
-  return /\.paypal\.(com|cn)(:\d+)?$/;
+  return /((\.paypal\.(com|cn))|(\.venmo\.com))(:\d+)?$/;
```
wsbrunson, line 65: I have a few suggestions/paths forward.

I think for this repo specifically, we should consider a `isVenmoDomain` function and a `isPayPalOwnedDomain` or `isPayPalTrustedDomain` function that combines `isPayPal` and `isVenmo`

If we start adding more domains like Honey, Braintree, or future products it'll make a lot of sense.

I'm guessing you are making this change for Zoid in `checkout-components`. I think we should really consider a new Zoid component for communicating with Venmo Web

ravishekhar: I like the idea of using a new method `isPayPalTrustedDomain()`. This will keep the method scalable to add more domains.

ravishekhar:
Closing in favor of https://github.com/paypal/paypal-sdk-client/pull/155

PR #152, opened 2023-05-11
chore: add load time metric

dtjones404:


src/tracking.js
```diff
@@ -123,10 +125,13 @@ export function setupLogger() {
         }`
       )
       .info(cache)
```
gregjopa, line 127: Can we also add this metric to the cal log?

```js
 .info(cache, { loadTime })
```

src/tracking.js
```diff
@@ -113,6 +113,8 @@ export function setupLogger() {
     // https://www.paypal.com/buttons/smart
     // https://www.paypal.com/us/gifts/
     const isLoadedInFrame = isPayPalDomain() && window.xprops;
+    const sdkLoadTime =
```
gregjopa, line 116: Nit: If we don't need the specific number type check, we can use this syntax to return a string value or undefined.
```suggestion
    const sdkLoadTime = loadTime && String(loadTime.toString);
```

dtjones404: Ahh load time can be zero though, which breaks the typing

gregjopa: oh good point. I forgot about falsy zero

gregjopa (approved):

PR #155, opened 2023-06-09
feat(venmo): domain support

mnicpt:
Provide domain support for Venmo in native flow allowing its domains to be trusted by PayPal.
https://paypal.atlassian.net/browse/DTCHKINT-1567
https://paypal.atlassian.net/browse/DTCHKINT-1535

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,5 +21,6 @@ module.exports = {
     __FUNDING_ELIGIBILITY__: true,
     __PAYPAL_DOMAIN__: true,
     __PAYPAL_API_DOMAIN__: true,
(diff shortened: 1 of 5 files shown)


nbierdeman (commented):
Nice work! Thanks, @mnicpt! I think adding a unit test for `isPayPalTrustedDomain()` would get us the code coverage we need to pass CI.

nbierdeman:
@mnicpt, something like this might do the trick: https://github.com/paypal/paypal-sdk-client/commit/b06077d7c4337e496730d2daec88baf708f92d8b

wsbrunson (approved):

PR #156, opened 2023-06-12
DTPPSDK-1216: Add cookie disabling flag to logger call 🍪

spencersablan:
As part of our cookie remediation effort, this PR will receive the `__DISABLE_SET_COOKIE__` global variable from the SDK, & use it to determine if the `logger` endpoint should send a query parameter (`disable-set-cookie=true`) to disable the non-essential cookies it sets.


wsbrunson (approved):

src/logger.js
```diff
@@ -3,12 +3,17 @@
 import { Logger, type LoggerType } from "@krakenjs/beaver-logger/src";
 import { memoize } from "@krakenjs/belter/src";
 
+import { getDisableSetCookie } from "./global";
 import { getPayPalLoggerUrl } from "./domains";
 
 type GetLogger = () => LoggerType;
 
+const loggerUrl = getDisableSetCookie()
+  ? `${getPayPalLoggerUrl()}?disable-set-cookie=true`
```
gregjopa, line 12: Originally I thought this would work but now I'm second guessing it. Can we pass query params like this on the logger HTTP Post request? 🤔

spencersablan: We can pass it! I was able to test this out. 😄 As long as Stargate can receive this (which they should be able to), we should be good: 
image

gregjopa: Great to hear! Thanks for testing this locally and ensuring it works. I'm cool with this solution 💯

wsbrunson (approved):

PR #157, opened 2023-06-15
fix(venmo): domain regex

mnicpt:
Fixed domain regex.

### DIFF
--- a/test/client/domains.js
+++ b/test/client/domains.js
@@ -70,7 +70,7 @@ describe(`domains test`, () => {
 
   it("should successfully match valid venmo testing domain", () => {
     window.__ENV__ = "local";
-    const validDomains = ["localhost.venmo.com"];
+    const validDomains = ["https://localhost.venmo.com"];
 
     for (const domain of validDomains) {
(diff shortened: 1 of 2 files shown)


test/client/domains.js
```diff
@@ -51,12 +51,14 @@ describe(`domains test`, () => {
 
   it("should successfully match valid venmo domain", () => {
     const validDomains = [
-      "www.venmo.com",
-      "www.venmo.com:8000",
-      "account.qa.venmo.com",
-      "www.account.qa.venmo.com",
-      "venmo.com",
-      "id.venmo.com",
+      "https://venmo.com",
```
wsbrunson, line 54: can we update the negative cases too
`"should not match invalid venmo domains`

mnicpt: Those are still good. It's just that we didn't expect these correctly. Needed to have protocol and port because that's what is required when checking for x-domain.

mnicpt: I'll add some anyway

wsbrunson (approved):

PR #163, opened 2023-08-03
[DTPPSDK-1319]Add getExperimentation()

xuanwan05:
Got
```
 __EXPERIMENTATION__: {
    __EXPERIENCE__: "1234, 4321";
    __TREATMENT__: "8765,7890";
  }
```
from global, then inside getExperimentation(), format it and return
```
experimation:{
    experience: "1234, 4321",
    treatment: "8765,7890"
}
```

### DIFF
--- a/src/declarations.js
+++ b/src/declarations.js
@@ -33,3 +33,6 @@ declare var __FUNDING_ELIGIBILITY__: FundingEligibilityType;
 
 // Experiment Variable
 declare var __DISABLE_SET_COOKIE__: boolean;
+declare var __EXPERIENCE__: string;
+declare var __TREATMENT__: string;
+declare var __EXPERIMENTATION__: Object;
--- a/src/declarations.js
+++ b/src/declarations.js
@@ -33,6 +33,7 @@ declare var __FUNDING_ELIGIBILITY__: FundingEligibilityType;
 
 // Experiment Variable
 declare var __DISABLE_SET_COOKIE__: boolean;
-declare var __EXPERIENCE__: string;
-declare var __TREATMENT__: string;
-declare var __EXPERIMENTATION__: Object;
+declare var __EXPERIMENTATION__: {|
+  __EXPERIENCE__?: string,
+  __TREATMENT__?: string,
+|};
--- a/src/global.js
+++ b/src/global.js
@@ -122,7 +122,7 @@ export function getDisableSetCookie(): boolean {
 
 export function getExperimentation(): Object {
   if (typeof __EXPERIMENTATION__ !== "undefined") {
-    let experimation = {};
+    const experimation = {};
     if (__EXPERIMENTATION__.__EXPERIENCE__) {
       experimation.experience = __EXPERIMENTATION__.__EXPERIENCE__;
     }
--- a/src/global.js
+++ b/src/global.js
@@ -119,3 +119,11 @@ export function getDisableSetCookie(): boolean {
 
   return false;
 }
+
+export function getExperimentation(): Object {
+  if (typeof __EXPERIMENTATION__ !== "undefined") {
+    return __EXPERIMENTATION__;
+  }
+
+  return null;
+}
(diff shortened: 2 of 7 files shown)


spencersablan (changes requested):
Hey Xuan! For this PR, we should only need the `getExperimentation()` function. We will not need the changes to logger 😅

src/declarations.js
```diff
@@ -33,3 +33,6 @@ declare var __FUNDING_ELIGIBILITY__: FundingEligibilityType;
 
 // Experiment Variable
 declare var __DISABLE_SET_COOKIE__: boolean;
+declare var __EXPERIENCE__: string;
+declare var __TREATMENT__: string;
+declare var __EXPERIMENTATION__: Object;
```
spencersablan, line 38: Can we just add a type called `Experimentation` that looks like this?
```
{
    __EXPERIENCE__: string;
    __TREATMENT__: string;
}

Comment (spencersablan) [reply src/declarations.js:38]: Then we only need to have this in the declaration file:
```
declare var __EXPERIMENTATION__: Experimentation;
```

& we can delete lines 36 & 37.

File: src/global.js [hunk 2]
```diff
@@ -119,3 +119,22 @@ export function getDisableSetCookie(): boolean {
...
+      if (__EXPERIMENTATION__.__TREATMENT__) {
+        experimation.treatment = __EXPERIMENTATION__.__TREATMENT__;
+      }
+      return experimation;
+    }
+
+    return null;
+  }
+
+  return null;
```
spencersablan, line 139: If there are no experiments, then `__EXPERIMENTATION__` should either be `undefined` or look like this:
```
{
    experience: "";
    treatment: "";
}
```

So I think we could adjust this a little bit:
```suggestion
export function getExperimentation(): Experimentation  {
  if (typeof __EXPERIMENTATION__ !== "undefined") {
      return __EXPERIMENTATION__;
  }
  
  return null;
```

test/client/global.js
```diff
@@ -307,4 +308,52 @@ describe(`globals cases`, () => {
...
+    const result = getExperimentation();
+
+    if (JSON.stringify(result) !== JSON.stringify(expectedResult)) {
+      throw new Error(
+        `Expected experimation to be ${JSON.stringify(
+          expectedResult
+        )}, got ${JSON.stringify(result)}`
+      );
+    }
+  });
```
spencersablan, line 358: Love that you wrote tests for this 🔥

src/global.js
```diff
@@ -119,3 +121,20 @@ export function getDisableSetCookie(): boolean {
 
   return false;
 }
+
+export function getExperimentation(): Experimentation | null {
+  if (typeof __EXPERIMENTATION__ !== "undefined") {
```
ravishekhar, line 126: Should we ensure this method always returns a value when `__EXPERIMENTATION__` is falsy e.g.
```json5
{
    experience: "",
    treatment: ""
}
```
This would avoid the clients of this method to do any null checks.

src/global.js
```diff
@@ -119,3 +121,16 @@ export function getDisableSetCookie(): boolean {
 
   return false;
 }
+
+export function getExperimentation(): Experimentation {
+  const experimation = {
+    experience: "",
+    treatment: "",
+  };
+  if (typeof __EXPERIMENTATION__ !== "undefined") {
```
ravishekhar, line 130: I like the idea of using optional chaining inside this block. Do we need this need the undefined check?

xuanwan05: If you are check `__EXPERIMENTATION__` and it's not defined, you will get undefined error. so we have to add this type check to guaranteed `__EXPERIMENTATION__` is defined.

ravishekhar: Ah ! thanks for pointing out the webpack global part 😇

ravishekhar (approved):
💯

spencersablan (approved):

PR #165, opened 2023-09-15
Decode the sdk meta if url encoded meta is passed

ravishekhar:
Decode SDK meta if the meta is passed as URL encoded value.

### DIFF
--- a/src/tracking.js
+++ b/src/tracking.js
@@ -114,7 +114,7 @@ export function setupLogger() {
     // https://www.paypal.com/us/gifts/
     const isLoadedInFrame = isPayPalDomain() && window.xprops;
     const sdkLoadTime =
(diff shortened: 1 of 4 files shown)


src/tracking.js
```diff
@@ -114,7 +114,7 @@ export function setupLogger() {
     // https://www.paypal.com/us/gifts/
     const isLoadedInFrame = isPayPalDomain() && window.xprops;
     const sdkLoadTime =
-      typeof loadTime === "number" ? loadTime.toString() : undefined;
+      typeof loadTime === "number" ? Number(loadTime) : undefined;
```
wsbrunson, line 117: isn't loadTime already a number in this case? since `typeof loadtime === "numer"` is true?

wsbrunson (approved):

wsbrunson (approved):

PR #170, opened 2023-11-27
feat(fraudnet): add fraudnet util for shared SDK components

siddy2181:
In order to support the latest pattern of pre-loading the fraudnet script and only collecting at the appropriate buyer intent trigger, we want to have the `loadFraudnet` steps available from sdk-client. This will load the script to any component that wants it/invokes `loadFraudnet` then provide them with a `collect` function to call at the point of buyer intent.

Initially this is in service of the new Connect component, but the intent is to bring it to the other integrations over time.


src/fraudnet.js
```diff
@@ -0,0 +1,100 @@
...
+  ({
+    env,
+    clientMetadataID,
+    cspNonce,
+    timeout = 1000,
+    appName,
+    queryStringParams = {},
+  }) => {
+    createConfigScript({ env, cspNonce, clientMetadataID, appName });
+    createFraudnetScript({ cspNonce, env, timeout, queryStringParams });
```
wsbrunson, line 98: do we need await either of these or return a promise?

wsbrunson:
it might be better to stick with jest and then we can come back and convert the whole repo to vite

src/constants.js
```diff
@@ -13,3 +14,15 @@ export const FPTI_TRANSITION = {
 export const FPTI_STATE = {
   PXP: ("PXP_CHECK": "PXP_CHECK"),
 };
+
+export const FRAUDNET_URL = {
```
wsbrunson, line 18: this is fine for our deadline but we should create a jira ticket to move these urls to csnw. I can help with the details

wsbrunson (approved):

cgdibble:
> it might be better to stick with jest and then we can come back and convert the whole repo to vite

Yup, we did just that actually. We pulled the changes off to a different branch to at least capture our efforts and cleaned this one and verified the changes manually.

cgdibble (approved):

PR #172, opened 2023-12-04
Add SDK token for PayPal domain

elizabethmv:
Restricting to PayPal domains for now, we will remove/add merchant domains in phases for Alpha/GA

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -37,7 +37,6 @@ import {
 import { getPath, getDefaultNamespace, getSDKHost } from "./global";
 import { CLIENT_ID_ALIAS } from "./config";
 import { getComputedLocales } from "./utils";
-import { isPayPalDomain } from "./domains";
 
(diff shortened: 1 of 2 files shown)


wsbrunson (approved):

src/script.js
```diff
@@ -338,6 +339,10 @@ export function getUserExperienceFlow(): ?string {
   return getSDKAttribute(SDK_SETTINGS.USER_EXPERIENCE_FLOW);
 }
 
+export function getSDKToken(): ?string {
+  return isPayPalDomain() ? getSDKAttribute(SDK_SETTINGS.SDK_TOKEN) : "";
```
wsbrunson, line 343: we should actually remove this check so that the axo project can start using this value too. They are about to go live with a merchant

elizabethmv: Makes sense, I'll get this removed

wsbrunson (approved):

PR #176, opened 2023-12-11
pass cmid as session id

wsbrunson:


src/session.js
```diff
@@ -8,6 +8,7 @@ import { getNamespace, getSDKAttribute } from "./script";
 function getSDKStorage(): Storage {
   return getStorage({
     name: getNamespace(),
+    stickySessionId: getClientMetadataID(),
```
spencersablan, line 11: Seeing this lint error right here, but I'm assuming we don't want to move `getClientMetadataID()` above this since that would breakup the exported functions.   🤷

spencersablan (approved):

PR #178, opened 2023-12-19
Add sendCountMetric to logger

elizabethmv:
This PR adds an exact copy of `sendCountMetric` here centrally, it was initially added in SPB but is being used in SDK repos. After this is in we can remove the one temporarily added in checkout-components

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -22,11 +22,11 @@ export const getLogger: GetLogger = memoize(() => {
 
 export const sendCountMetric = ({
   dimensions,
-  event = "unused",
+  event,
   name,
   value = 1,
 }: {|
-  event?: string,
+  event: string,
   name: string,
   value?: number,
   dimensions: {


src/logger.js
```diff
@@ -19,3 +19,25 @@ export const getLogger: GetLogger = memoize(() => {
     url: loggerUrl,
   });
 });
+
+export const sendCountMetric = ({
+  dimensions,
+  event = "unused",
```
wsbrunson, line 25: we should make event required, i was wrong when i initially set these as `unused`. It makes the queries for metrics really long if they don't have an event to filter on

elizabethmv: Makes sense, checking to see if there are any being used without event

wsbrunson (approved):

src/logger.js
```diff
@@ -19,3 +19,25 @@ export const getLogger: GetLogger = memoize(() => {
     url: loggerUrl,
   });
 });
+
+export const sendCountMetric = ({
```
okorye, line 23:
```suggestion
export const sendCountMetric = ({
```

PR #179, opened 2023-12-19
feature:DTPPCPSDK-1509-Add Shopper Insights Usage into sdkMeta

prashantjagasia:
When creating the sdk meta, we will need to check whether the shopper insights api was used. This value will be set in session state.

### DIFF
--- a/src/meta.js
+++ b/src/meta.js
@@ -4,6 +4,7 @@ import { SDK_SETTINGS } from "@paypal/sdk-constants/src";
 import { base64encode, ATTRIBUTES } from "@krakenjs/belter/src";
 
 import { getScriptUrl, getSDKAttributes } from "./script";
+import { getSessionState } from "./session";
 
 const ALLOWED_ATTRS = [
   SDK_SETTINGS.AMOUNT,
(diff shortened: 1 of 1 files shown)


src/meta.js
```diff
@@ -21,12 +22,17 @@ export function getSDKMeta(): string {
   const url = getScriptUrl();
 
   const scriptAttrs = getSDKAttributes();
+  const handler = () => {};
+  const shopperInsights = getSessionState(handler);
```
wsbrunson, line 26:
```suggestion
  const shopperInsightsUsed = getSessionState(state => state.shopperInsights.getRecommendedPaymentMethodsUsed)
```

https://github.com/paypal/paypal-checkout-components/pull/2286/files#diff-6d1caeaa7a4f194561101db7781e6ff562dd9a95c9cb07852bc69ce6443af15eR18

src/meta.js
```diff
@@ -21,12 +22,17 @@ export function getSDKMeta(): string {
...
+
   const attrs = {};
   for (const attr of Object.keys(scriptAttrs)) {
     if (ALLOWED_ATTRS.indexOf(attr) !== -1) {
       attrs[attr] = scriptAttrs[attr];
     }
   }
-
-  return base64encode(JSON.stringify({ url, attrs })).replace(/\=+$/, ""); // eslint-disable-line no-useless-escape
+  return base64encode(JSON.stringify({ url, attrs, shopperInsights })).replace(
```
wsbrunson, line 34:
```suggestion
  const sdkMeta = {
    url,
    attrs,
    features: {
      shopperInsightsUsed
    },
  }
  return base64encode(JSON.stringify(sdkMeta)).replace(
```

PR #180, opened 2023-12-27
DTPPCPSDK-1511 CCO ShopperInsights

prashantjagasia:
We’ll need to send down sShopper insights information from session to smart-payment-buttons.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -34,7 +34,6 @@ import {
   PAGE_TYPES,
 } from "@paypal/sdk-constants/src";
 
-import { getSessionState } from "./session"; // eslint-disable-line import/no-cycle
 import { getPath, getDefaultNamespace, getSDKHost } from "./global";
 import { CLIENT_ID_ALIAS } from "./config";
 import { getComputedLocales } from "./utils";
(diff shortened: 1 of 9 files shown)


src/script.js
```diff
@@ -34,6 +34,7 @@ import {
   PAGE_TYPES,
 } from "@paypal/sdk-constants/src";
 
+import { getSessionState } from "./session"; // eslint-disable-line import/no-cycle
```
ravishekhar, line 37: Lets move this logic to an another module or we can even introduce a new module to avoid the cyclic dependency between script and session.

src/script.js
```diff
@@ -368,3 +369,9 @@ export function getCountry(): $Values<typeof COUNTRY> {
 export function getLang(): $Values<typeof LANG> {
   return getLocale().lang;
 }
+
+export function getShopperInsights(): boolean {
```
wsbrunson, line 373: since this is pulling from session state, we should move it to `src/session.js`

wsbrunson, line 373: we should name this `getShopperInsightsUsed` or `getRecommendedPaymentMethodsUsed`

wsbrunson (approved):

PR #117, opened 2022-02-28
Increase the test coverage on the client to avoid failing checks in the CI

borodovisin:
### Description
The PR contains a new set of unit tests to increase the client(Karma) code coverage.

### Why are we making these changes?
To avoid the failing CI check related to the client unit tests. Now is set to a minimum of 50% on the Codecov config. Previously we had a code coverage under the desire 50% of coverage and CI check fails for all the PRs.

### Note
The name of the test cases probably aren't accurate, any feedback on that is welcome.

PR #181, opened 2023-12-29
chore: migrate to vitest

cgdibble:
This does a lot.

- sets up for use of `vitest` in the repo
- remove Kraken testing libs, mocha, jest, and karma 
- adopts idiomatic `vitest` patterns of mocking and assertions
- removes cases of test-aware implementations (some still linger on, I addressed as I ran in to them with fixing tests)
- removes logger tests that really test beaver logger internals
- removes more complicated mock script util that had it's own test suite
- merge `script` and `scriptUtils` tests since those were all tests of `script.js`
- uses `vitests` coverage as well. Dumps all coverage in to a single coverage dir since it's no longer split between test runners. Merges coverage uploading to a single GH Action step, too.
- removes old test tool files and references from package.json scripts
- co-locates test files with implementation and ditches the ./test dir entirely

This spawned from a need to test recent fraudnet changes. So be mindful that this base branch is to our fraudnet feature branch, and that we still need to _add_ tests for fraudnet so that is still pending.

Note: Initially done here in #117 but that was using a feature branch as the base branch. This just moves the same changes to PR against `main`.


src/fraudnet.js
```diff
@@ -68,7 +64,8 @@ export const createConfigScript = ({
     configScript.setAttribute("fncls", FRAUDNET_FNCLS);
     configScript.text = JSON.stringify(config);
     // eslint-disable-next-line compat/compat
-    document.body?.appendChild(configScript);
+    document.body.appendChild(configScript);
```
cgdibble, line 67:
```suggestion
    document.body?.appendChild(configScript);
```

src/domains.js
```diff
@@ -16,16 +16,6 @@ import {
 import { URI } from "./config";
 
 export function getPayPalLoggerDomain(): string {
-  if (__ENV__ === ENV.LOCAL) {
```
cgdibble, line 19: This is an indirect way to be test-aware. We shouldn't be changing how we go about things like this. Environmental differences happen, but this is just avoiding the real approach done in `getPayPalDomain` for a simpler test

src/script.js
```diff
@@ -49,18 +49,6 @@ const buildScriptNotFoundError = (host, path, error) => {
 };
 
 export const getSDKScript: GetSDKScript = memoize(() => {
-  if (__TEST__) {
```
cgdibble, line 52: Test aware implementations make me a 😿

src/index.js
```diff
@@ -12,7 +12,6 @@ export * from "./api";
 export * from "./experiment";
 export * from "./session";
 export * from "./events";
-export * from "./test";
```
cgdibble, line 15: TODO:
This file was specifically just for creating a mock SDK. It lived in the implementation. 
One thing to worry about here is if `SDK-client` was exporting this so other repositories could use it. That might require us to at least keep it in here for now since I don't think we can pin those other repos to a different version of SDK client.

test/client/api.test.js
```diff
@@ -0,0 +1,139 @@
...
+    app_id: "APP-80W284485P519543T",
+    expires_in: 31838,
+    nonce: "2022-03-07T22:41:38ZqHkiC0_odfzFwo27_X0wVuF67STYq39KRplBeeyY2bk",
+    error: null,
+  };
+
+  beforeEach(() => {
+    memoize.clear();
+    window.__PAYPAL_DOMAIN__ = "testurl";
+    getCurrentScript.mockReturnValue({
```
cgdibble, line 36: TODO: there was a lot of futzing with replacing the `insertMockSDKScript` or whatever it was. But now that I've gone through all this, I want to circle back and refresh on how to make a simple util. It could just be at the vitest setup level if we wanted, too.

test/client/domains.test.js
```diff
@@ -0,0 +1,106 @@
...
+      "www.paypal.cn.example.com",
+    ];
+
+    for (const domain of invalidDomains) {
+      expect(domain).not.toMatch(getPayPalDomainRegex());
+    }
+  });
+
+  it("should successfully match valid venmo domain", () => {
+    const validDomains = [
```
cgdibble, line 48: I would expect these would be something to come from a constant, but felt it was a creeping scope to chase that down, too.

test/client/domains.test.js
```diff
@@ -0,0 +1,106 @@
...
+    ];
+
+    for (const domain of validDomains) {
+      expect(domain).toMatch(getVenmoDomainRegex());
+    }
+  });
+
+  it("should successfully match valid venmo testing domain", () => {
+    window.__ENV__ = "local";
+    const validDomains = ["https://localhost.venmo.com"];
```
cgdibble, line 66: TODO: fix this to not iterate as no iteration is needed for...one

cgdibble: Using ```window.__ENV__ = "local"``` is likely making another test-aware implementation switch here. But, I didn't want to chase down all of these at once as it fans out a lot it seems.

vite.config.js
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable spaced-comment */
+// eslint-disable-next-line @typescript-eslint/triple-slash-reference
+/// <reference types="vitest" />
```
cgdibble, line 3: TODO: clean up some of this copy-pasta commented out stuff

test/client/api.test.js
```diff
@@ -0,0 +1,139 @@
+/* @flow */
```
jshawl, line 1: are you open to moving these right next to the implementation? so:

```
src/
  api.js
  api.test.js
```

instead of:

```
src/
  api.js
test/
  api.test.js
```

just for consistency with paypal-checkout-components and smart-payment-buttons?

vite.config.js
```diff
@@ -0,0 +1,55 @@
...
+      include: ["src/**/*.js"],
+    },
+  },
+  optimizeDeps: {
+    esbuildOptions: {
+      plugins: [esbuildFlowPlugin()],
+    },
+  },
+  plugins: [flowPlugin({ exclude: "" })],
+});
```
jshawl, line 55: I had this thought:

> we should create a new grumbler-scripts-esque repo to contain all the vite config and npm scripts

but I think I prefer how it is now. e.g. it would take like 3 repo hops to identify where the karma dependency is coming from https://github.com/paypal/paypal-sdk-client/pull/181/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L14

just sharing the idea in case it's worth pursuing further (still on team independent repo config fwiw)

package.json
```diff
@@ -21,9 +20,10 @@
     "setup": "npm install && npm run flow-typed",
     "format": "prettier --write --ignore-unknown .",
     "format:check": "prettier --check .",
-    "test": "npm run format:check && npm run lint && npm run flow-typed && npm run flow && npm run jest && npm run karma",
+    "test": "npm run format:check && npm run lint && npm run flow-typed && npm run flow && npm run test:unit",
     "webpack": "babel-node --plugins=transform-es2015-modules-commonjs ./node_modules/.bin/webpack --progress",
-    "jest": "jest test/server --env=node --no-cache --collectCoverageFrom='server/' --coverageDirectory='coverage/jest' --coverage --verbose --runInBand --silent=false",
+    "test:unit:watch": "vitest",
+    "test:unit": "vitest run --coverage",
```
jshawl, line 26: it would be :100: to standardize w/ the other repos:

paypal checkout components has https://github.com/paypal/paypal-checkout-components/blob/635ab8b162a4ce9a8fb116bdcfc9e4a8edfdd116/package.json#L32C5-L32C31

smart payment buttons has https://github.com/paypal/paypal-smart-payment-buttons/blob/main/package.json#L54 and https://github.com/paypal/paypal-smart-payment-buttons/blob/7b28401d2463c52e7da564e76097690119fde2f7/package.json#L33 but also in vite config https://github.com/paypal/paypal-smart-payment-buttons/blob/main/vite.config.js#L43-L46

not sure what all the differences are but pointing out in case you wanted sdk-client to be the gold standard! 🏅

vite.config.js
westeezy: we can just add those config to grumbler-scripts vs creating a new repo. it might not be a perfect tool (grumbler-scripts) but it would be consistent to keep shared configs there

westeezy: i always hated how grumbler-scripts was intended to hide dependencies as it does. thats also something we can pursue moving away from and I'd be in big support but lets stay incremental so we keep progress moving

src/domains.js
westeezy: This might change behavior in stage though unfortunately. Do you know if stage will still be the proper url? Stage is considered local if memory serves, which is because locally they had still logged to msmaster endpoints.

test/client/api.test.js
westeezy: I really like colocated tests. Would be potentially easier to consider this an incremental change (fast follow) for another PR pending agreement on this one so far. But would be cool with seeing it all in one PR too, whichever is easiest for author.

src/domains.js
wsbrunson: or local is considered stage. by default, in the local env we send all requests to the stage url unless you override it with a proxy

src/fraudnet.js
```diff
@@ -91,7 +88,6 @@ export const createFraudnetScript = ({
 
     fraudnetScript.setAttribute("nonce", cspNonce || "");
     fraudnetScript.setAttribute("src", fraudnetUrl);
-    fraudnetScript.addEventListener("error", () => resolve());
```
wsbrunson, line 94: wouldn't we still want this? in case there's an error loading the script, otherwise it would stall out forever?

test/client/api.test.js
wsbrunson: this was our agreed upon structure moving forward: https://github.com/paypal/paypal-smart-payment-buttons/tree/main/test

package.json
cgdibble: Yeah I LOVE when they all match. Reducing the amount of cognitive load needed for any given repo switch is important to me. 

`vitest run` is needed because vitest defaults to watch mode and `run` disables `watch`. This is used for and needed by CI so it doesn't sit open. To minimize impact to existing paradigms, I went with adding `test:unit:watch`.

`npm test` is a bit overloaded IMO, but didn't want to mess with that in this scope.

src/domains.js
cgdibble: ohhhh interesting, ok then I can add this back!! False assumption/understanding on my part!

cgdibble: Is there a reason we handle it so differently when it is stage?

src/fraudnet.js
cgdibble: You are correct, thank you. I think this snuck in with a merge conflict or something. I'll update.

cgdibble: Oh oops nevermind it's actually still there but this one was a duplicate to one below where we actually handle the error within this fn

src/index.js
cgdibble: Ugh actually looks like it is being used: https://github.com/search?q=org%3Apaypal%20insertMockSDKScript&type=code

I'll restore that file, then.

test/client/api.test.js
cgdibble: I am _absolutely_ a fan of co-locating the tests. That should be a small change so I'll timebox it for this PR.

cgdibble: Thanks for suggesting this @jshawl, I've updated it accordingly!

vite.config.js
cgdibble: I'm intrigued by the idea, but do want to consider that out of the scope of this PR. A shared config could help a lot of the set up for sure.

src/session.test.js
```diff
@@ -0,0 +1,73 @@
...
+  getStorageState,
+  getStorageID,
+  getSessionState,
+  getClientMetadataID,
+} from "./session";
+
+const clientId = "foobar123";
+const mockScriptSrc = `https://test.paypal.com/sdk/js?client-id=${clientId}`;
+
+function makeMockScriptElement(src = mockScriptSrc): HTMLCanvasElement {
```
cgdibble, line 15: I did this in a few files because it's smaller scope and more understandable/manageable (IMO) than the `insertMockSDK` or whatever fn in src/test.js

But, it's also repetitive. I could put it in a file like `./test/helpers`? Any opinions?

jshawl (approved):
Nice work! :shipit:

src/session.js
```diff
@@ -12,7 +12,6 @@ export function getClientMetadataID(): ?string {
 function getSDKStorage(): Storage {
   return getStorage({
     name: getNamespace(),
-    stickySessionId: getClientMetadataID() || "",
```
cgdibble, line 15: This was showing as a Flow error as getStorage doesn't have a param for `stickySessionId`. I do have a worry that I'm missing something with how our build system might _actually_ make it so it does lol @wsbrunson am I missing anything in that regard?

wsbrunson: do you have the latest version of `belter`?

src/domains.js
```diff
@@ -8,10 +8,10 @@ import {
 } from "@krakenjs/cross-domain-utils/src";
 
 import {
-  getProtocol,
-  getStageHost,
   getPayPalDomain,
   getPayPalAPIDomain,
+  getStageHost,
```
SancSalix, line 13: totally a nit but if this order was changed, why not alphabetical?

src/session.js
cgdibble: Oh weird, I do but I think when I initially did this Belter was outa date. But I've updated this now, thanks!

src/domains.js
cgdibble: Fixed!

PR #184, opened 2024-02-15
add DPoP encoding utils

jshawl:
This PR implements a few utility functions that will be used in an upcoming PR for jwt creation and signatures as part of the larger [DPoP](https://datatracker.ietf.org/doc/html/rfc9449) initiative that will be used in the hosted buttons component. 

I am planning on two follow up prs to this repo:

- one that implements the key pair generation and signing and verification
- one that implements the jwt creation and exports a `dPoPHeaders` function that will be used in checkout-components.

### DIFF
--- a/src/dpop.js
+++ b/src/dpop.js
@@ -1,6 +1,5 @@
 /* @flow */
-
-import { ZalgoPromise } from "@krakenjs/zalgo-promise/src";
+/* eslint-disable promise/no-native, no-restricted-globals */
 
 export const stringToBytes = (string: string): Uint8Array => {
   // eslint-disable-next-line compat/compat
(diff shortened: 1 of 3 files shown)


src/dpop.js
```diff
@@ -0,0 +1,33 @@
...
+    .replace(/\//g, "_");
+};
+
+export const base64decodeUrlSafe = (string: string): string => {
+  return atob(string.replace(/-/g, "+").replace(/_/g, "/"));
+};
+
+export const sha256 = (string: string): ZalgoPromise<string> => {
+  const bytes = stringToBytes(string);
+  return window.crypto.subtle.digest("sha-256", bytes).then((digest) => {
```
wsbrunson, line 28: You could make this a regular promise, since this `.then` would negate any of the synchronous abilities of Zalgo.

wsbrunson: and if so, i'd suggest going with `async/await`

wsbrunson (approved):

src/dpop.js
jshawl: amazing! would you `eslint-disable` these?

```
24:47  error  Unexpected use of 'Promise'  no-restricted-globals
24:47  error  "Promise" is not defined     promise/no-native
```

it looks like it is explicitly enabled here - https://github.com/krakenjs/grumbler-scripts/blame/3d70c3c2a809793b420fadf4c6859016fdcbb094/config/.eslintrc-browser.js#L8-L9 🤔

wsbrunson: yeah i usually disable them at the top of the file

PR #185, opened 2024-02-15
use new metric helpers

wsbrunson:
* use the new `metricCounter` interface
* move one of the metrics outside of an if statement that doesn't apply to the metric
* add base metric namespace for ppcp and the dimensions for v5 web.
* add deprecated comment to `sendMetricCount`

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -2,6 +2,7 @@
 
 import { Logger, type LoggerType } from "@krakenjs/beaver-logger/src";
 import { memoize } from "@krakenjs/belter/src";
+import { BASE_SDK_METRIC_NAMESPACE } from "@paypal/sdk-constants/src";
 
 import { getDisableSetCookie } from "./global";
 import { getPayPalLoggerUrl } from "./domains";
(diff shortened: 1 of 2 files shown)


src/tracking.js
```diff
@@ -181,20 +183,6 @@ export function setupLogger() {
         //     token: getTokenType(),
         //   },
         // })
-        // $FlowIssue
```
jshawl, line 184: type safety ++ !

src/logger.js
```diff
@@ -17,9 +18,11 @@ export const getLogger: GetLogger = memoize(() => {
 
   return Logger({
     url: loggerUrl,
+    metricNamespacePrefix: BASE_SDK_METRIC_NAMESPACE,
   });
 });
 
+// Deprecated - will remove once we have verified no other clients are importing this function
```
jshawl, line 25: option to use jsoc:

```
/** 
 * @deprecated - will remove...
 */
```

and then the message will show on hover in other code bases.

jshawl (approved):
Looks good! one non-blocking comment 🚀

src/logger.js
wsbrunson: oh wow great idea!

PR #189, opened 2024-02-22
use node version from .nvmrc in ci

jshawl:
This PR fixes an issue where pull requests are tested using node 16 but npm publish actions are tested using node 18.

I copied over the yaml from `publish.yml`:
https://github.com/paypal/paypal-sdk-client/blob/4b87e823d890abbc7187dfa0800277e237427a59/.github/workflows/publish.yml#L14-L18


wsbrunson (approved):
thank you!

PR #192, opened 2024-03-14
Fix for incorrect metric causing high cardinality

sioked:
The `isLoadedInFrame` variable previously was `isPayPalDomain() && window.xprops`. Because of the && operator, the value here would be the value of window.xprops which is a very high cardinality value. We ended up having metrics disabled due to the cardinality of these events so casting to a Boolean should solve for that.

One thing I'm not clear on is why we're mapping `isLoadedInFrame` to `isPayPalDomain` and not calling `isPayPalDomain()` directly, but I'm only trying to solve the cardinality issue on this event.

### DIFF
--- a/src/tracking.js
+++ b/src/tracking.js
@@ -171,26 +171,11 @@ export function setupLogger() {
     );
 
     if (loadTime) {
-      logger
-        // We can not send gauge metrics to our logger backend currently
-        // once we have that ability, we should uncomment this gauge metric
-        // .metricGauge({
-        //   namespace: "sdk_client.init.gauge",
-        //   event: "load_performance",
-        //   value: sdkLoadTime,
-        //   dimensions: {
-        //     cacheType,
-        //     version,
-        //     components: getComponents().join(","),
-        //     isPayPalDomain: isLoadedInFrame,
-        //     token: getTokenType(),
-        //   },
-        // })
-        .track({
-          [FPTI_KEY.TRANSITION]: "process_js_sdk_init_client",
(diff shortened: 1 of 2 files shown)


src/tracking.js
```diff
@@ -203,7 +188,7 @@ export function setupLogger() {
       dimensions: {
         components: getComponents().join(","),
         integrationSource,
-        isPayPalDomain: isLoadedInFrame,
+        isPayPalDomain: Boolean(isLoadedInFrame),
```
elizabethmv, line 191: I'm not sure if this is still the case but we've seen in the past signalFX dropping boolean values. Can we make sure this is a true/false string or some other string value?

jshawl (approved):
I like the change (after the lint fix) but wanted to propose an optional/additional change in beaver-logger around https://github.com/krakenjs/beaver-logger/blob/87749c48788dfd2525fb9b075ca33a7bf69d3311/src/logger.js#L296-L301

that limits the max cardinality value (truncate ~and `console.warn`~ maybe?)

src/tracking.js
sioked: Good observation! I wasn't aware of that - I'll change this to a string

elizabethmv (approved):

sioked:
@jshawl I would love to find a way to validate/limit the cardinality in beaver logger. We could look at the length of the string as a possible indicator, but it is possible to have low cardinality across all of the metrics while also having a long string length. Do you have any thoughts or ideas?

PR #193, opened 2024-03-15
Hardcode library values

nbierdeman:
### Description

@gregjopa, @westeezy, and I had a solid chat about DTPPCPSDK-1482 where it was decided we'd remove `JS_SDK_LIBRARIES` from `@paypal/sdk-constants` and hardcode values. This was inspired by the difficulty of adding `@paypal/sdk-constants` as a dependency of `paypal-js`. Additionally, it was decided that `"none"` was a better value than `"raw-script"` for when a library isn't being used to load the JS SDK.

### DIFF
--- a/src/script.test.js
+++ b/src/script.test.js
@@ -6,6 +6,7 @@ import { SDK_SETTINGS } from "@paypal/sdk-constants/src";
 
 import { makeMockScriptElement } from "../test/helpers";
 
+import { getJsLibrary } from "./tracking";
 import {
   getClientID,
   getIntent,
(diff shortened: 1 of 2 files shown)


src/script.test.js
```diff
@@ -255,14 +255,22 @@ describe(`script cases`, () => {
...
   it("should successfully get js sdk library", () => {
-    const jsSdkLibrary = JS_SDK_LIBRARIES.REACT_PAYPAL_JS;
+    const jsSdkLibrary = "react-paypal-js";
     const mockElement = makeMockScriptElement(mockScriptSrc);
     mockElement.setAttribute(SDK_SETTINGS.JS_SDK_LIBRARY, jsSdkLibrary);
     // $FlowIgnore
     getCurrentScript.mockReturnValue(mockElement);
 
-    expect(getJsSdkLibrary()).toEqual(jsSdkLibrary);
+    expect(getJsLibrary()).toEqual(jsSdkLibrary);
```
jshawl, line 273: nice!

jshawl (approved):
:shipit:

PR #194, opened 2024-04-17
Add initial entry point for vault flows with SDK token

elizabethmv:
This PR is an initial entrance to allow the latest vault flow (ID token) to work the new SDK token. It will not enable the e2e flow just yet.

When SDK token is passed alone, internally we will set the ID token as the SDK token value. We also will use this new getCustomerId function that decodes the SDK token for UX improvements to prevent an API call/correct loading state.

### DIFF
--- a/src/script.js
+++ b/src/script.js
@@ -235,10 +235,6 @@ export function getAmount(): ?string {
   return amount;
 }
 
-export function getUserIDToken(): ?string {
-  return getSDKAttribute(SDK_SETTINGS.USER_ID_TOKEN);
-}
-
 export function getClientAccessToken(): ?string {
   const clientToken = getClientToken();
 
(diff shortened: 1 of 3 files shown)


src/script.js
```diff
@@ -330,10 +326,47 @@ export function getUserExperienceFlow(): ?string {
...
 
+type decodedCustomerId = (string) => string;
+export const decodeCustomerIdFromToken: decodedCustomerId = memoize((token) => {
+  try {
+    if (token && typeof atob === "function") {
+      const { options = {} } = JSON.parse(window.atob(token.split(".")[1]));
+      return options.customer_id || "";
+    }
+
+    return "";
```
ravishekhar, line 359: Should we return a null in case there is no customer id?

ravishekhar (approved):
left a minor comment, looks good otherwise.

src/script.test.js
```diff
@@ -546,6 +548,52 @@ describe(`script cases`, () => {
...
+    const mockElement = makeMockScriptElement(mockScriptSrc);
+    mockElement.setAttribute("data-sdk-client-token", inputToken);
+    mockElement.setAttribute("data-user-id-token", inputToken);
+    // $FlowIgnore
+    getCurrentScript.mockReturnValue(mockElement);
+
+    expect(getSDKToken).toThrow("Do not pass SDK token and ID token");
+  });
+
+  it("getCustomerId returns a string of the decoded customer_id from the SDK token", () => {
```
wsbrunson, line 584: consider adding the negative case where there is no token or there is no customer_id in the token (or both)

wsbrunson (approved):

src/script.test.js
elizabethmv: Good call, done 👍

PR #199, opened 2024-05-03
fix invalid venmo domain tests

ravishekhar:
The test case written to validate invalid Venmo domains are currently asserting for invalid domain by not passing protocol (http/https) in the input values.
This results in `www.venmo.com` being validated as invalid Venmo domain, and `https://www.venmo.com` being valid domain. 

A crucial part of invalid domain validation must be to also validate the eTLD+1 of the Venmo domain. 

This PR fixes by adding the protocol to the input values.

### DIFF
--- a/src/domains.test.js
+++ b/src/domains.test.js
@@ -65,9 +65,16 @@ describe(`domains test`, () => {
 
   it("should not match invalid venmo domains", () => {
     const invalidDomains = [
-      "www.venmo.com.example.com",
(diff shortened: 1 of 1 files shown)


jshawl (approved):
🚀

wsbrunson (approved):

elizabethmv (approved):

spencersablan (approved):

PR #201, opened 2024-05-17
Updated validation of the web-sdk bridge URL.

bywood:
This updates the validation for the NextGen SDK bridge URL. The merchant `origin` is no longer being passed as a parameter on the Checkout URL, but will be included in the `sdkMeta` as a parameter on the bridge URL. The bridge URL will also include a `version` parameter.


gregjopa (approved):
Great work @bywood! 💯

wsbrunson (approved):

PR #202, opened 2024-05-21
feat: add compatibility with window alias strategy

dtjones404:
[DTPPCPSDK-2163](https://paypal.atlassian.net/browse/DTPPCPSDK-2163)

### DIFF
--- a/src/setup.js
+++ b/src/setup.js
@@ -4,6 +4,7 @@ import { destroyElement } from "@krakenjs/belter/src";
 
 import { getVersion } from "./global";
 import { getSDKScript, getNamespace } from "./script";
+import { getLogger } from "./logger";
 
 export type SetupComponent<T> = {|
   name: string,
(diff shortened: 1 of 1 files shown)


jshawl (approved):
nice!

src/setup.js
```diff
@@ -21,7 +22,13 @@ export function setupSDK(components: $ReadOnlyArray<SetupComponent<mixed>>) {
   const existingVersion = existingNamespace && existingNamespace.version;
 
   if (existingNamespace) {
-    if (existingNamespace[INTERNAL_DESTROY_KEY]) {
+    if (existingVersion.startsWith("6.")) {
```
bywood, line 25: Is there a scenario where a merchant's code could be using v6 on `window.paypal`, but then this changes it to `window.paypal.v6` and breaks their code?

dtjones404: Correct, I think the tradeoff for v6 defaulting to `window.paypal` is that we have to teach merchants using both SDKs to access v6 through `window.paypal.v6`.

bywood (approved):

PR #204, opened 2024-05-24
Add wasShopperInsightsUsed 👾

spencersablan:
**JIRA:** [DTPPCPSDK-2369](https://paypal.atlassian.net/browse/DTPPCPSDK-2369)

- _**💥 Adds wasShopperInsightsUsed function**_ - Checks session state for key indicators that shopper insights was used

### DIFF
--- a/src/session.test.js
+++ b/src/session.test.js
@@ -77,7 +77,58 @@ describe("session cases", () => {
     });
   });
 
-  it("wasShopperInsightsUsed should return false if shopperInsights doesn't exist in the state", () => {
-    expect(wasShopperInsightsUsed()).toEqual(false);
+  describe.only("wasShopperInsightsUsed", () => {
+    let mockState;
+
+    const mockGetSessionState = (handler: (state: Object) => void) => {
+      handler(mockState);
+    };
+
+    it("should return false if shopperInsights doesn't exist in the state", () => {
+      expect(wasShopperInsightsUsed()).toEqual(false);
+    });
+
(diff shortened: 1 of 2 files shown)


src/session.test.js
```diff
@@ -75,4 +76,8 @@ describe("session cases", () => {
       stickySessionId: expect.any(String),
     });
   });
+
+  it("wasShopperInsightsUsed should return false if shopperInsights doesn't exist in the state", () => {
```
wsbrunson, line 80: we should a true case as well

wsbrunson (approved):

nikrom17 (approved):

src/session.test.js
spencersablan: Definitely. 👍   I was having some issues with mocking some of the storage functions before my vacation & realized all the other functions only really have negative testing.  But let me see if a fresh brain can come up with something better now 😄

spencersablan: Haha actually, it was pretty easy. See.. fresh brain 🧠

wsbrunson: fresh brains solve most things

src/session.test.js
```diff
@@ -75,4 +76,59 @@ describe("session cases", () => {
       stickySessionId: expect.any(String),
     });
   });
+
+  describe.only("wasShopperInsightsUsed", () => {
```
wsbrunson, line 80:
```suggestion
  describe("wasShopperInsightsUsed", () => {
```

wsbrunson (approved):

PR #206, opened 2024-05-29
add getFirstRenderExperiments

nikrom17:
## Description
This PR is to support first render experiments
First render experiments can be added to runtime globals without messing with the caching strategy via the build globals.

## Jira
https://paypal.atlassian.net/browse/DTPPCPSDK-2355


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@paypal/sdk-client",
-  "version": "4.0.189",
+  "version": "4.0.188-alpha.0",
```
wsbrunson, line 3: we can back this out

src/global.js
```diff
@@ -122,19 +122,9 @@ export function getDisableSetCookie(): boolean {
   return false;
 }
 
-export function getExperimentation(): Experimentation | null {
```
wsbrunson, line 125: you may want to leave this function as is until you can remove it from the other components that are using it.

Unless its not being used right now

nikrom17: sounds good, I don't think it is being used but I'd rather keep it than find out via deployment errors.

I'll add a `// TODO first-render-experiment-cleanup <description>` comment in all the places we should revisit

imbrian (approved):

src/global.js
```diff
@@ -122,19 +122,27 @@ export function getDisableSetCookie(): boolean {
...
   }
 
   return null;
 }
+
+export function getFirstRenderExperiments(): FirstRenderExperiments | null {
+  if (typeof __FIRST_RENDER_EXPERIMENTS__ !== "undefined") {
+    return __FIRST_RENDER_EXPERIMENTS__;
+  }
+  return null;
```
wsbrunson, line 147: could this return `{}` so that the code using it won't have to null check?

They'll have to see if their key was included regardless, this would take away checking if the object was null too

wsbrunson (approved):

PR #213, opened 2024-06-18
keep v4 patch versions at 132 or above

wsbrunson:
Description in slack


server/meta.test.js
```diff
@@ -11,363 +11,211 @@ afterEach(() => {
...
+      expected: "https://www.paypalobjects.com/api/checkout.4.0.132.js",
+    },
+    {
+      windowName: "xcomponent__ppcheckout__4_0_132__abc12345",
+      expected: "https://www.paypalobjects.com/api/checkout.4.0.132.js",
+    },
+    {
+      windowName: "xcomponent__ppcheckout__4_0_133__abc12345",
+      expected: "https://www.paypalobjects.com/api/checkout.4.0.133.js",
+    },
```
wsbrunson, line 89: Here are the tests that matter

server/meta.jsx
```diff
@@ -360,6 +360,21 @@ export function unpackSDKMeta(sdkMeta?: string): SDKMeta {
...
+                            // if patch version is less than 132, we want to 
+                            // set the version to 132. If for some reason we can't
+                            // parse out a patch version, set version as latest
+                            // if neither cases are true, leave version alone because
+                            // it can be more than a semver version number ("min" for example)
+                            if (/4\\.0\\.\\d{1,3}/.test(version)) {
+                              var patchString = version?.split('.')?.pop()
+
+                              if (!patchString) {
+                                version = 'latest'
```
cspalevic, line 372: above on L359, if we see `latest` then we set `version` to an empty string. Should this be an empty string here instead?

wsbrunson: oh damn i had that backwards, thank you.

I need to add a test case for if patch version is undefined, i think i know how

wsbrunson: okay, don't think its possible to have an invalid patch version but i'm leaving that there just in case

cspalevic (approved):

imbrian (approved):

wsbrunson:
i just realized we want to wait to merge this until we make the edge routing changes: https://github.com/paypal/paypal-sdk-client/pull/216

PR #217, opened 2024-09-05
Expose utility functions for custom script builder usage

andrewatwood:
Expose internal utilities for use cases where inserting the raw `script` tag provided by `getSDKLoader` is not ideal:

- `getSDKScriptUrl` to get the `src` of the output `script` tag if present
- `getSDKScriptAttributes` to get any `data-*` attributes set on the output `script` tag if present
- `getSDKInlineScript` to get the inline script populating the `script` tag if present

### DIFF
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -24,8 +24,15 @@ import {
 } from "./constants";
 import { constHas, entries } from "./util";
 
+type SDKAttributes = {|
+  [string]: string | boolean,
+|};
+
 type SDKMeta = {|
   getSDKLoader: (options?: {| baseURL?: string, nonce?: string |}) => string,
+  getSDKScriptAttributes: () => SDKAttributes | void,
+  getSDKScriptUrl: () => string | void,
+  getSDKInlineScript: (baseURL?: string) => string | void,
 |};
 
 const emailRegex = /^.+@.+$/;
@@ -239,10 +246,6 @@ function validateSDKUrl(sdkUrl: string) {
   }
 }
 
-type SDKAttributes = {|
-  [string]: string | boolean,
-|};
-
 const getDefaultSDKAttributes = (): SDKAttributes => {
   // $FlowFixMe
   return {};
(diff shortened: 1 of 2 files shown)


server/meta.test.js
```diff
@@ -1399,3 +1399,67 @@ test("should error when the origin parameter is not just the origin", () => {
     throw new Error("Expected error to be thrown");
   }
 });
+
+describe("Utilities", () => {
```
andrewatwood, line 1403: Rather than checking every utility every time, I figured we could just check that they're equal and go from there. But it might not hurt to add specific assertions for each utility to each test case.

I started there, but it ended up being so many cases that it seemed redundant.

server/meta.jsx
```diff
@@ -19,8 +19,15 @@ import {
 } from "./constants";
 import { constHas, entries } from "./util";
 
+type SDKAttributes = {|
+  [string]: string | boolean,
+|};
+
 type SDKMeta = {|
   getSDKLoader: (options?: {| baseURL?: string, nonce?: string |}) => string,
+  getSDKScriptAttributes: () => SDKAttributes | void,
```
gregjopa, line 28: If this will only return data attributes then I think the name should reflect that.
```suggestion
  getSDKScriptDataAttributes: () => SDKAttributes | void,
```

spencersablan: good point greg

bywood (commented):
We recently updated the web-sdk bridge validation and the data attribute validation (https://github.com/paypal/paypal-sdk-client/pull/289). This was done to facilitate making changes to the `sdkMeta` structure in order to support additional web-sdk bridge scripts. So currently, we're relying on `jsx-pragmatic`'s `render(html())` method to prevent XSS attacks. Since these new methods bypass that rendering, we'll need to make sure that what is returned is properly sanitized.

andrewatwood:
Nice, glad this will go in! I mentioned this on Slack too but feel free to make changes directly if you beat me to the punch, all sounds good to me. It may be a little later today before I get availability to do so.

PR #218, opened 2024-10-18
Add support for 'payment-flow' query param for the web-sdk bridge

gregjopa:
This PR adds support for the new `payment-flow` query parameter for the web-sdk bridge url.

### DIFF
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -91,17 +91,31 @@ function validateWebSDKUrl({ pathname, query }) {
     );
   }
   // check for extraneous parameters
-  Object.keys(query).forEach((param) => {
-    if (param !== "version" && param !== "origin") {
+  const validWebSDKBridgeParams = ["origin", "version", "payment-flow"];
+  for (const param of Object.keys(query)) {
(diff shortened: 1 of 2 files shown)


wsbrunson (approved):

server/meta.jsx
```diff
@@ -91,17 +91,31 @@ function validateWebSDKUrl({ pathname, query }) {
     );
   }
   // check for extraneous parameters
-  Object.keys(query).forEach((param) => {
-    if (param !== "version" && param !== "origin") {
+  const validWebSDKBridgeParams = ["origin", "version", "payment-flow"];
```
dtjones404, line 94: should we add 'debug' here?

gregjopa: Good point @dtjones404! I'll send a follow up PR for adding support for the optional `debug` param.

gregjopa: Good point @dtjones404! I'll send a follow up PR for adding support for the optional `debug` param.

gregjopa: @dtjones404 here's the follow up PR to add support for the `debug` query param: https://github.com/paypal/paypal-sdk-client/pull/219

PR #220, opened 2024-10-21
chore: allow extra parameters to pass through to the bridge url

gregjopa:
This PR updates the web sdk bridge logic to allow for extra parameters to be passed through. These extra parameters will be validated and accepted/rejected at a different step outside this codebase. The purpose of this change is to make updating sdk-client easier in the future for the web sdk.

### DIFF
--- a/server/meta.jsx
+++ b/server/meta.jsx
@@ -90,58 +90,44 @@ function validateWebSDKUrl({ pathname, query }) {
       `Invalid path for web-sdk bridge url: ${pathname || "undefined"}`
     );
   }
-  // check for extraneous parameters
-  const validWebSDKBridgeParams = [
-    "origin",
-    "version",
-    "payment-flow",
-    "debug",
-  ];
-  for (const param of Object.keys(query)) {
-    if (!validWebSDKBridgeParams.includes(param)) {
-      throw new Error(`Invalid parameter on web-sdk bridge url: ${param}`);
-    }
-  }
+
+  // supported query string parameters
+  const { origin, version, ["payment-flow"]: paymentflow, debug } = query;
 
   // validate the version parameter
-  if (query.version === undefined || !semverRegex.test(query.version)) {
+  if (version === undefined || !semverRegex.test(version)) {
     throw new Error(
-      `Invalid version parameter on web-sdk bridge url: ${query.version}`
+      `Invalid version parameter on web-sdk bridge url: ${version}`
     );
   }
 
   // validate the payment-flow parameter
   const validPaymentFlows = ["popup", "modal", "payment-handler"];
-  if (
-    query["payment-flow"] === undefined ||
-    !validPaymentFlows.includes(query["payment-flow"])
-  ) {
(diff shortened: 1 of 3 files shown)


server/meta.test.js
```diff
@@ -1251,28 +1251,35 @@ test("should construct a valid web-sdk bridge url", () => {
...
+  const src = script.attr("src");
+  const uid = script.attr("data-uid");
+
+  console.log(sdkUrl);
+
+  if (src !== sdkUrl) {
+    throw new Error(`Expected script url to be ${sdkUrl} - got ${src}`);
+  }
+  if (uid !== sdkUID) {
+    throw new Error(`Expected data UID be ${sdkUID} - got ${uid}`);
```
wsbrunson, line 1282: i was about to say "you should use an expect for this" but i just realized what repo we are in 😂

wsbrunson (approved):

server/meta.jsx
```diff
@@ -90,58 +90,44 @@ function validateWebSDKUrl({ pathname, query }) {
...
-    "debug",
-  ];
-  for (const param of Object.keys(query)) {
-    if (!validWebSDKBridgeParams.includes(param)) {
-      throw new Error(`Invalid parameter on web-sdk bridge url: ${param}`);
-    }
-  }
+
+  // supported query string parameters
+  const { origin, version, ["payment-flow"]: paymentflow, debug } = query;
```
bywood, line 95: Do we still want or need to maintain the list of required parameters?

bywood: I suppose this allows whoever is using `unpackSDKMeta` to catch these errors, whereas it would be harder to detect that SDK server responded with a validation error.

server/meta.jsx
```diff
@@ -90,58 +90,44 @@ function validateWebSDKUrl({ pathname, query }) {
...
     );
   }
 
   // validate the payment-flow parameter
   const validPaymentFlows = ["popup", "modal", "payment-handler"];
-  if (
-    query["payment-flow"] === undefined ||
-    !validPaymentFlows.includes(query["payment-flow"])
-  ) {
+  if (paymentflow === undefined || !validPaymentFlows.includes(paymentflow)) {
```
dtjones404, line 106: Would it be better to do this validation on the server-side as well?

bywood (approved):

server/meta.jsx
gregjopa: When would this line of code error? I may be missing something. Best I can tell, `query` should always be an object. And if one of these values is not defined, it would not error:
```js
const query = {};
const { origin, version } = query;
console.log(origin) // => undefined
```

I added this destructuring just for readability. The code below it contains the checks for if the values are required or not (ex: `if (version === undefined`);

server/meta.jsx
gregjopa: I'm open to removing it here.

gregjopa:
We chatted through this PR. Going to close this out as unmerged.

PR #224, opened 2025-01-28
feat: support new method to validate url

ravishekhar:
- Support a new method `isPayPalTrustedUrl()` which validates if the provided URL is PayPal or Venmo URL. 
- Using the standard URL object instead of `getActualDomain` for future compatibility https://github.com/krakenjs/cross-domain-utils/blob/8f7bb2d9e07c465a5c625724b980f1f551dd28fa/src/utils.js#L91

### DIFF
--- a/src/domains.js
+++ b/src/domains.js
@@ -80,3 +80,17 @@ export function isPayPalTrustedDomain(): boolean {
     Boolean(getDomain().match(getVenmoDomainRegex()))
   );
 }
+
+export function isPayPalTrustedUrl(href: string): boolean {
+  try {
+    // eslint-disable-next-line compat/compat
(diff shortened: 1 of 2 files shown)


mchoun (commented):
I'm just curious where this new method is being utilized? Is it replacing all instances of `getActualDomain`?

ravishekhar:
> I'm just curious where this new method is being utilized? Is it replacing all instances of `getActualDomain`?

This method will be utilized in other modules like checkout-components to validate if a particular URL is safe to be redirected to.

mchoun (approved):

wsbrunson (approved):

977812671 (approved):

PR #228, opened 2025-08-06
Adding globalSession to maintain the global_paypal storage

gusveloso:
## What is the purpose of this PR?
Retrieve `globalSessionID` value from a new session created by Messaging components when more than a namespace are being used at the same time. Besides retrieving it, we should log the data on tracking button events on a prop named `global_session_id`

## Type of change 

- [x] New feature (backward compatible change that adds new capability).
- [ ] UI change
- [ ] Bug fix.
- [ ] Breaking change (backward incompatible change). Provide references to customer impact and communication.
- [ ] Refactor (no functional change)

## Testing Plan

Here's the step by step to check if this change is working, but both changes on `paypal-sdk-client`, `smartcomponentnodeweb`, `paypal-sdk-constants` and `paypal-checkout-components` are needed

- Open your site, network tab and wait for logger requests
- Wait for any button event like `process_button_load`
- Check if `global_session_id` property exists. IT SHOULDN'T
- After it, go to console, copy and paste this:
```
window.localStorage.__paypal_global_5_0_500_storage__ = JSON.stringify({
   globalSessionID: "uid_testID",
})
```
- Reload the page and check the same logger, the `global_session_id` should be present and with uid_testID as its value

### DIFF
--- a/src/globalSession.js
+++ b/src/globalSession.js
@@ -2,12 +2,19 @@
 
 import { getStorage, type Storage } from "@krakenjs/belter/src";
 import { ZalgoPromise } from "@krakenjs/zalgo-promise/src";
+
 import { getVersion } from "./global";
+import { getLogger } from "./logger";
 
 const GLOBAL_SESSION_ID_KEY = "globalSessionID";
 
 export function getGlobalSessionName(): string {
-  const sdkVersion = String(getVersion()).replace(/[\W]+/g, "_");
+  let version = getVersion();
+  if (!version) {
+    getLogger().warn("global_session_no_sdk_version");
+    version = "unknown";
+  }
+  const sdkVersion = String(version).replace(/[\W]+/g, "_");
   return `paypal_global_${sdkVersion}`;
 }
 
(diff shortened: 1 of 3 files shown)


src/globalSession.js
```diff
@@ -0,0 +1,34 @@
+/* @flow */
+
+import { getStorage, type Storage } from "@krakenjs/belter/src";
+import { ZalgoPromise } from "@krakenjs/zalgo-promise/src";
+import { getVersion } from "./global";
+
+const GLOBAL_SESSION_ID_KEY = "globalSessionID";
+
+export function getGlobalSessionName(): string {
+  const sdkVersion = String(getVersion()).replace(/[\W]+/g, "_");
```
nikrom17, line 10: What are the failure cases for the `getVersion` function?
if it doesn't find the version or if a user clears local storage manually, what will happen?

src/globalSession.js
```diff
@@ -19,15 +26,26 @@ function getGlobalStorageState<T>(handler: (storage: Object) => T): T {
...
-  );
-  return globalSessionID.value;
+export function getGlobalSessionID(): string | void {
+  const globalSessionID = getGlobalStorageState((state) => {
+    if (
+      !state ||
+      typeof state !== "object" ||
+      !state.hasOwnProperty(GLOBAL_SESSION_ID_KEY)
+    ) {
+      return undefined;
```
nikrom17, line 36: Lets add a log here for not finding the global session

nikrom17 (approved):

